PR #1, opened 2019-06-06
[WIP] Script to set up Windows node with Kubeadm

ksubrmnn:


kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,288 @@
...
+    Get-Help $bin -Detailed
+
+    $usage = "
+    Usage: 
+		$bin [-help] [-init] [-join] [-reset]
+
+	Examples:
+        $bin -help                                                           print this help
+        $bin -InstallPrerequisites -ConfigFile kubecluster.json               Set up this Windows node to run containers
+        $bin -init -ConfigFile kubecluster.json                              joins the windows node to existing cluster control-plane (unsupported)
```
mkostersitz: Looks like a copy paste error, init should not work on Windows right?

ksubrmnn: I said control-plane to say that this Windows node can't be a master and then added (unsupported). Kubeadm has `init` and also `join --experimental-control-plane` so I tried to cover both scenarios here. I can rephrase it to be more clear

kubeadm/v1.15.0/KubeClusterHelper.psm1
```diff
@@ -0,0 +1,1528 @@
...
+    else {
+        throw "$Global:InterfaceName doesn't exist"
+    }
+}
+
+function ValidateConfig()
+{
+    if ($Global:Cni -ne "flannel")
+    {
+        throw "$Global:Cni not yet supported"
```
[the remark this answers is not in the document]
ksubrmnn: $Global:Cni is loaded with $Global:ClusterConfiguration.Cni.Name which is the user input

kubeadm/v1.15.0/KubeClusterHelper.psm1
```diff
@@ -0,0 +1,1528 @@
...
+    }
+    
+    if ($Global:NetworkPlugin -ne "vxlan" -and $Global:NetworkPlugin -ne "bridge")
+    {
+        throw "$Global:NetworkPlugin is not yet supported"
```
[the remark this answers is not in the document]
ksubrmnn: $Global:NetworkPlugin is loaded with $Global:ClusterConfiguration.Cni.Plugin.Name

kubeadm/v1.15.0/KubeCluster.ps1
neolit123: i think the `-init` flag can be indeed confusing, possibly omit completely for now.

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,288 @@
...
+        $bin -join -ConfigFile kubecluster.json                              joins the windows node to existing cluster
```
neolit123: > join this Windows worker node to an existing cluster
[1]

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,288 @@
...
+        $bin -reset -ConfigFile kubecluster.json                             reset the kubernetes cluster
```
neolit123: > reset and cleanup this Windows worker node
[2]

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,288 @@
...
+# Licensed under the MIT license.
+#
+Param(
+    [parameter(Mandatory = $false,HelpMessage="Print the help")]
+    [switch] $help,
+    [parameter(Mandatory = $false,HelpMessage="Initialize Windows Control Plane node (unsupported)")]
+    [switch] $init,
+    [parameter(Mandatory = $false,HelpMessage="Install pre-requisites")]
+    [switch] $InstallPrerequisites,
+    [parameter(Mandatory = $false,HelpMessage="Join the windows node to the master")]
```
neolit123: [1]

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,288 @@
...
+    [switch] $join,
+    [parameter(Mandatory = $false,HelpMessage="Reset this windows node and cleanup everything")]
```
neolit123: [2]
is it possible to define these HelpMessages only once and use them in both the params and usage screen?

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,288 @@
...
+    [switch] $reset,
+    [parameter(Mandatory = $false,HelpMessage="Path to input configuration json ")]
```
neolit123: `json`->`JSON`

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,288 @@
...
+        DownloadKubeConfig -Master $Global:MasterIp -User $Global:MasterUsername
+        if (!(KubeConfigExists))
+        {
+            throw $kubeConfig + " does not exist. Cannot connect to the master cluster"
+        }
+    }
+
+    # Validate connectivity with Master API Server
+
+    Write-Host "Trying to connect to the Kubernetes master"
```
neolit123: just a FYI, usage of "control-plane node" is preferred over "master" in user facing text.
there were debates about this.

seems to apply to multiple locations in the scripts.

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,288 @@
...
+    $bin = $PSCommandPath 
+    Get-Help $bin -Detailed
+
+    $usage = "
+    Usage: 
+		$bin [-help] [-init] [-join] [-reset]
+
+	Examples:
+        $bin -help                                                           print this help
+        $bin -InstallPrerequisites -ConfigFile kubecluster.json               Set up this Windows node to run containers
```
neolit123: possibly call this flag `-install`?

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,288 @@
...
+)
+
+function Usage()
+{
+    $bin = $PSCommandPath 
+    Get-Help $bin -Detailed
+
+    $usage = "
+    Usage: 
+		$bin [-help] [-init] [-join] [-reset]
```
neolit123: `install..` is missing in the list

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,288 @@
...
+    wget $Path -o $tmpPath
+    ipmo $tmpPath  -DisableNameChecking
+    Remove-Item $tmpPath
+}
+###############################################################################################
+# Download pre-req scripts
+
+$helperPath = "https://raw.githubusercontent.com/kubernetes-sigs/sig-windows-tools/kubeadm/v1.15.0/KubeClusterHelper.psm1"
+$helperDestination = "$PSScriptRoot\KubeClusterHelper.psm1" 
+wget $helperPath -o $helperDestination
```
neolit123: has something installed wget already?
possibly use `Invoke-WebRequest` instead?

kubeadm/v1.15.0/KubeClusterHelper.psm1
```diff
@@ -0,0 +1,1528 @@
...
+        $output.Close()
+        $input.Close()
+    }
+}
+
+function Install-7Zip()
+{
+    # ask to install 7zip, if it's not already installed
+    if (-not (Get-Command Expand-7Zip -ErrorAction Ignore)) {
+        Install-Package -Scope CurrentUser -Force 7Zip4PowerShell -Verbose
```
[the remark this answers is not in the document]
ksubrmnn: This is only used to extract the Kubernetes binaries.

michmike:
@ksubrmnn have we made any updates to this to progress from WIP? thanks!

kubeadm/v1.15.0/KubeClusterHelper.psm1
```diff
@@ -0,0 +1,1521 @@
...
+    }
+}
+
+function DownloadAndExtractTarGz($url, $dstPath)
+{
+    $tmpTarGz = New-TemporaryFile | Rename-Item -NewName { $_ -replace 'tmp$', 'tar.gz' } -PassThru
+    $tmpTar = New-TemporaryFile | Rename-Item -NewName { $_ -replace 'tmp$', 'tar' } -PassThru
+    DownloadFile -Url $url -Destination $tmpTarGz.FullName -Force
+    #Invoke-WebRequest $url -o $tmpTarGz.FullName
+    Expand-GZip $tmpTarGz.FullName $tmpTar.FullName
```
benmoss: Why not just use the BSD tar that is included with Windows? https://techcommunity.microsoft.com/t5/Containers/Tar-and-Curl-Come-to-Windows/ba-p/382409

kubeadm/v1.15.0/KubeClusterHelper.psm1
```diff
@@ -0,0 +1,1528 @@
...
+    { 
+        if ($insecureProtocols -notcontains $protocol) 
+        { 
+            $secureProtocols += $protocol 
+        } 
+    } 
+    [System.Net.ServicePointManager]::SecurityProtocol = $secureProtocols
+    
+    try {
+        (New-Object System.Net.WebClient).DownloadFile($Url,$Destination)
```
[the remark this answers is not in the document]
benmoss: or `curl.exe` is my preferred downloader nowadays :)

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,280 @@
...
+{
+    $tmpPath = [io.Path]::Combine([System.IO.Path]::GetTempPath(), [io.path]::GetFileName($Path))
+    wget $Path -o $tmpPath
+    ipmo $tmpPath  -DisableNameChecking
+    Remove-Item $tmpPath
+}
+###############################################################################################
+# Download pre-req scripts
+
+$helperPath = "https://raw.githubusercontent.com/kubernetes-sigs/sig-windows-tools/kubeadm/v1.15.0/KubeClusterHelper.psm1"
```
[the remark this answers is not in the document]
benmoss: We have https://sigs.k8s.io, but I don't think you can link to a raw file that way. I would think that updating aliases would cause problems, I'd rather version these things together and force users to update the scripts together. We probably shouldn't even be reaching out to the internet for these things IMHO.

kubeadm/v1.15.0/KubeClusterHelper.psm1
```diff
@@ -0,0 +1,1521 @@
...
+    New-Service -Name "kubelet" -StartupType Automatic -BinaryPathName "$kubeletArgs"
+    kubeadm join "$(GetAPIServerEndpoint)" --token "$Global:Token" --discovery-token-ca-cert-hash "$Global:CAHash"
+    # Open firewall for 10250. Required for kubectl exec pod <>
+    if (!(Get-NetFirewallRule -Name KubeletAllow10250 -ErrorAction SilentlyContinue ))
+    {
+        New-NetFirewallRule -Name KubeletAllow10250 -Description "Kubelet Allow 10250" -Action Allow -LocalPort 10250 -Enabled True -DisplayName "KubeletAllow10250" -Protocol TCP -ErrorAction Stop
+    }
+    if (!(IsNodeRegistered)) {
+        Write-Host "$(IsNodeRegistered)"
```
benmoss: isn't this just going to `Write-Host $false`?

kubeadm/v1.15.0/KubeCluster.ps1
ksubrmnn: Unfortunately no, we need literals here. Constant variables don't work either.

ksubrmnn:
@jocelynberrendonner

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,273 @@
...
+                ServerCore = "mcr.microsoft.com/windows/servercore:ltsc2019";
+            }
+        }
+    }
+}
+function LoadPsm1($Path)
+{
+    $tmpPath = [io.Path]::Combine([System.IO.Path]::GetTempPath(), [io.path]::GetFileName($Path))
+    Invoke-WebRequest $Path -o $tmpPath
+    Import-Module $tmpPath  -DisableNameChecking
```
PatrickLang: can you check if this already exists before downloading it? When testing PRs like this it's easier if you can pull all the files down and test together.

```
if (!(Test-Path $tmpPath)) { invoke-webrequest... }
```

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,273 @@
...
+    Remove-Item $tmpPath
+}
+###############################################################################################
+# Download pre-req scripts
+
+$helperPath = "https://raw.githubusercontent.com/kubernetes-sigs/sig-windows-tools/kubeadm/v1.15.0/KubeClusterHelper.psm1"
+$helperDestination = "$PSScriptRoot\KubeClusterHelper.psm1" 
+Invoke-WebRequest $helperPath -o $helperDestination
```
PatrickLang: same here - if local use it for easier testing

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,273 @@
...
+    Write-Host "Service CIDR    : $ServiceCIDR"
+    Write-Host "DNS ServiceIp   : $KubeDnsServiceIP"
+    Write-Host "############################################"
+
+    #
+    # Install Services & Start in the below order
+    # 1. Install & Start Kubelet
+    InstallKubelet  -CniDir $(GetCniPath) `
+                -CniConf $(GetCniConfigPath) -KubeDnsServiceIp $KubeDnsServiceIp `
+                -NodeIp $Global:ManagementIp -KubeletFeatureGates $KubeletFeatureGates
```
PatrickLang: Where's InstallKubelet implemented? It needs to set resolv-conf="" on Windows, otherwise you get an error starting the pods such as:

```

Events:
  Type     Reason                  Age               From               Message
  ----     ------                  ----              ----               -------
  Normal   Scheduled               24s               default-scheduler  Successfully assigned default/iis-2019-5cd8f558dc-xjxhl to win1
  Warning  FailedCreatePodSandBox  9s (x2 over 23s)  kubelet, win1      Failed create pod sandbox: open /etc/resolv.conf: The system cannot find the file specified.
```

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,273 @@
...
+    [parameter(Mandatory = $false,HelpMessage="Reset and clean up this Windows worker node")]
+    [switch] $reset,
+    [parameter(Mandatory = $false,HelpMessage="Path to input configuration JSON")] 
+    $ConfigFile
+)
+
+# Handle --help
+if ($help.IsPresent)
+{
+    Usage
```
PatrickLang: this needs to go below Usage() otherwise `./kubecluster.ps1 -help` throws an error saying Usage is not defined

kubeadm/v1.15.0/Kubeclusterbridge.json
```diff
@@ -0,0 +1,44 @@
...
+            "Name": "bridge"
+        },
+        "InterfaceName" : "Ethernet"
+    },
+    "Kubernetes" : {
+        "Source" : {
+            "Release" : "1.14.0",
+            "Url" : "https://dl.k8s.io/v1.14.1/kubernetes-node-windows-amd64.tar.gz"
+        },
+        "Master" : {
```
PatrickLang: This needs to be renamed to ControlPlane. The script expects Kubernetes.ControlPlane

kubeadm/v1.15.0/Kubeclustervxlan.json
```diff
@@ -0,0 +1,44 @@
...
+            "Name": "vxlan"
+        },
+        "InterfaceName" : "Ethernet"
+    },
+    "Kubernetes" : {
+        "Source" : {
+            "Release" : "1.14.0",
+            "Url" : "https://dl.k8s.io/v1.14.0/kubernetes-node-windows-amd64.tar.gz"
+        },
+        "Master" : {
```
PatrickLang: This needs to be renamed `ControlPlane` to pass `if (!$Global:ClusterConfiguration.Kubernetes.ControlPlane)` in the script

kubeadm/v1.15.0/Kubeclusterbridge.json
```diff
@@ -0,0 +1,44 @@
...
+        ],
+        "Plugin" : {
+            "Name": "bridge"
+        },
+        "InterfaceName" : "Ethernet"
+    },
+    "Kubernetes" : {
+        "Source" : {
+            "Release" : "1.14.0",
+            "Url" : "https://dl.k8s.io/v1.14.1/kubernetes-node-windows-amd64.tar.gz"
```
PatrickLang: this line and above - 1.15 now :)

kubeadm/v1.15.0/Kubeclustervxlan.json
```diff
@@ -0,0 +1,44 @@
+{
+    "Cri" : {
+        "Name" : "dockerd",
+        "Images" : {
+            "Pause" : "mcr.microsoft.com/k8s/core/pause:1.0.0",
```
PatrickLang: 1.2.0 supports kubectl port-forward

kubeadm/v1.15.0/Kubeclusterbridge.json
```diff
@@ -0,0 +1,44 @@
+{
+    "Cri" : {
+        "Name" : "dockerd",
+        "Images" : {
+            "Pause" : "mcr.microsoft.com/k8s/core/pause:1.0.0",
```
PatrickLang: 1.2.0 is better to support `kubectl port-forward`

PatrickLang:
I recommended a few changes. The main problem I'm having right now though is that the kubelet is still trying to map /etc/resolv.conf into the pods. that's the default on Linux but it causes pods to fail to start on Windows. We need to fix that in the script before we update docs to use it.

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,273 @@
...
+    Write-Host "############################################"
+    Write-Host "Able to connect to the control-plane node"
+    Write-Host "Discovered the following"
+    Write-Host "Cluster CIDR    : $ClusterCIDR"
+    Write-Host "Service CIDR    : $ServiceCIDR"
+    Write-Host "DNS ServiceIp   : $KubeDnsServiceIP"
+    Write-Host "############################################"
+
+    #
+    # Install Services & Start in the below order
```
daschott: After a possible reboot, there is no guarantee the services will start up again in the right order, or?
Is it possible to add dependency chain using sc after registering the services? E.g.
`sc config ServiceA depend= ServiceB`

kubeadm/v1.15.0/KubeCluster.ps1
daschott: I think the --resolv-conf parameter is set in the config file now, right?

kubeadm/v1.15.0/KubeCluster.ps1
ksubrmnn: Kube-Proxy and Flannel don't need to be started in a particular order anymore.

@PatrickLang Does it matter what order kubelet and kubeproxy start on reboot?

daschott: If it helps, I tried validating launching kube-proxy and kubelet in reverse order on a new WS2019 machine against an existing cluster and didn't notice any difference

ksubrmnn: Okay, FlannelD can be started before Kubelet as well. So I think we don't need to create dependent services. However, reboot isn't officially supported yet.

daschott: True, I should've said "crashes" instead of "reboot" -  I'll be sure to document that reboot isn't supported yet as well so that expectations are clear 👍

kubeadm/v1.15.0/KubeClusterHelper.psm1
```diff
@@ -0,0 +1,1515 @@
...
+        $nodeName = (hostname).ToLower()
+        CreateService -ServiceName FlannelD -CommandLine $flanneldArgs `
+            -LogFile "$log" -EnvVaribles @{NODE_NAME = "$nodeName";}    
+    }
+}
+
+function UnInstallFlannelD()
+{
+    RemoveService -ServiceName FlannelD
+    Remove-Item $(GetKubeFlannelPath) -Force -ErrorAction SilentlyContinue
```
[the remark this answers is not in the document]
PatrickLang: I think that's ok because there could be other configs there that weren't created by this script. That directory can be left behind on Linux. For example, you might uninstall the kubelet but have static manifests left in /etc/kubernetes . Then you install a new version and use those files again.

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,273 @@
...
+    Write-Host "touch ~/.ssh/authorized_keys"
+    Write-Host "echo $pubKey >> ~/.ssh/authorized_keys"
+
+    $res = Read-Host "Continue to Reboot the host [Y/n] - Default [Y] : "
+    if ($res -eq '' -or $res -eq 'Y'  -or $res -eq 'y')
+    {
+        Restart-Computer -Force
+    }
+
+    InstallCRI $Global:Cri
```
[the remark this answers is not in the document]
PatrickLang: Yeah, I started those manually. Is a reboot actually needed?

kubeadm/v1.15.0/KubeClusterHelper.psm1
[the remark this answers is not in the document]
ksubrmnn: This way is the most compatible across different Windows SKUs

kubeadm/v1.15.0/KubeCluster.ps1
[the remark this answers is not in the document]
ksubrmnn: Reboot is needed if we have to install the container role. I can download the binaries first and then check the container role and reboot if needed. Regardless a new shell is needed for ssh.

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,273 @@
...
+if ($help.IsPresent)
+{
+    Usage
+    exit
+}
+
+function ReadKubeclusterConfig($ConfigFile)
+{
+    # Read the configuration and initialize default values if not found
+    $Global:ClusterConfiguration = ConvertFrom-Json ((GetFileContent $ConfigFile -ErrorAction Stop) | out-string)
```
mkostersitz: Would it be better to use Try/Catch to tell the user what is wrong with the JSON? rather than relying on them to be able to decipher the PowerShell error?

kubeadm/v1.15.0/KubeCluster.ps1
mkostersitz: In the case of simply checking if the file exists and skipping pulling it down we would need to do a version check on the file incase the one on disk is older than the one upstream. I would suggest always pull down for simplicity. And yes pulling everything at the start makes more sense than doing them one by one. Ideally pull a tar ball or ZIP file and unpack it locally, much more efficient IMHO

kubeadm/v1.15.0/KubeClusterHelper.psm1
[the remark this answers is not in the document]
mkostersitz: Are you not using Invoke-WebRequest most other places?

kubeadm/v1.15.0/KubeCluster.ps1
ksubrmnn: Added it to the new-service kubelet call

kubeadm/v1.15.0/KubeClusterHelper.psm1
ksubrmnn: Done

kubeadm/v1.15.0/KubeClusterHelper.psm1
ksubrmnn: Removed

kubeadm/v1.15.0/KubeCluster.ps1
ksubrmnn: Switched to using DownloadFile which will check if it already exists. Per offline discussions with Patrick and m2, we're continuing with using Test-Path since it'll allow for developers to more easily use their custom files

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,280 @@
...
+        Restart-Computer -Force
+    }
+
+    exit
+}
+
+# Handle -Join
+if ($Join.IsPresent)
+{
+    $kubeConfig = GetKubeConfig
```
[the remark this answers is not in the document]
ksubrmnn: GetKubeConfig just returns the path so it's being used as the destination for scp and for logging messages

kubeadm/v1.15.0/KubeClusterHelper.psm1
```diff
@@ -0,0 +1,1528 @@
...
+$Global:NetworkMode = "l2bridge"
+$Global:DockerImageTag = "1809"
+$Global:Configuration = @{}
+$Global:MasterUsername = "localadmin"
+$Global:PauseImage = "mcr.microsoft.com/k8s/core/pause:1.0.0"
+$Global:NanoserverImage = "mcr.microsoft.com/windows/nanoserver:1809"
+$Global:ServercoreImage = "mcr.microsoft.com/windows/servercore:ltsc2019"
+$Global:Cri = "dockerd"
+
+function GetKubeConfig()
```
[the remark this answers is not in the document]
ksubrmnn: It's just used to resolve a path, much like the other Get* functions. This keeps things decoupled

kubeadm/v1.15.0/KubeCluster.ps1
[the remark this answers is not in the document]
ksubrmnn: Resuming script after restart in Restart-And-Run function

kubeadm/v1.15.0/KubeClusterHelper.psm1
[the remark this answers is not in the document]
ksubrmnn: It leaves behind an empty ` C:\etc\kubernetes\pki` which seems to follow the kubeadm output of `Deleting contents of config directories: [\etc\kubernetes\manifests /etc/kubernetes/pki]`

ksubrmnn:
@PatrickLang @daschott The script should be in good shape now. Only change is to pull cni binaries from the containernetworkingplugins repo. I'll add that change tomorrow but otherwise it looks good!

@daschott can you try this with l2bridge and also run the validation script? I haven't run the validation script yet for Overlay but I'll do that in the morning :)

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,321 @@
...
+    {
+        $res = Read-Host "Do you wish to generate a SSH Key & Add it to the Linux control-plane node [Y/n] - Default [Y]"
+        if ($res -eq '' -or $res -eq 'Y'  -or $res -eq 'y')
+        {
+            ssh-keygen.exe
+        }
+    }
+
+    $pubKey = Get-Content $env:HOMEDRIVE/$env:HOMEPATH/.ssh/id_rsa.pub
+    Write-Host "Execute the below commands on the Linux control-plane node ($Global:MasterIp) to add this Windows node's public key to its authorized keys"
```
daschott: Can we move the ssh stuff to line 216 at the end of the install procedure? The output from `InstallCRI` overwrites this txt quickly and there are no pauses so users may overlook that this is an action they must take :)

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    Write-Output "Creating scheduled task action ($scriptPath $argList)..."
+    $action = New-ScheduledTaskAction -Execute "powershell.exe" -Argument "-NoExit $scriptPath $argList"
+
+    Write-Output "Creating scheduled task trigger..."
+    $trigger = New-ScheduledTaskTrigger -AtLogOn
+
+    Write-Output "Registering script to re-run at next user logon..."
+    Register-ScheduledTask -TaskName $global:BootstrapTask -Action $action -Trigger $trigger -RunLevel Highest | Out-Null
```
daschott: This is failing for me. Where is $global:BootstrapTask set?

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -0,0 +1,321 @@
...
+
+    try
+    {
+        if ($Force)
```
daschott: Where is $Force set?

kubeadm/v1.15.0/KubeClusterHelper.psm1
```diff
@@ -0,0 +1,1360 @@
...
+        WriteKubeClusterConfig
+    }
+}
+
+function InstallDockerImages()
+{
+    if (!(docker images $Global:NanoserverImage -q))
+    {
+        docker pull $Global:NanoserverImage
+        if (!$LastExitCode) {
```
daschott: This is failing for me, despite successful docker image pull. Should this be inverted to either
`if ($LastExitCode)` or `if (!($LastExitCode -eq 0))` ?

kubeadm/v1.15.0/KubeClusterHelper.psm1
```diff
@@ -0,0 +1,1360 @@
...
+            throw "Failed to pull $Global:NanoserverImage"
+        }
+    }
+    docker tag $Global:NanoserverImage mcr.microsoft.com/windows/nanoserver:latest
+    if (!(docker images $Global:ServercoreImage -q))
+    {
+        docker pull $Global:ServercoreImage
+        if (!$LastExitCode) {
```
daschott: This is failing for me, despite successful docker image pull. Should this be inverted to either
if ($LastExitCode) or if (!($LastExitCode -eq 0)) ?

kubeadm/v1.15.0/Kubeclustervxlan.json
```diff
@@ -0,0 +1,44 @@
...
+            "Release" : "1.15.0",
+            "Url" : "https://dl.k8s.io/v1.15.0/kubernetes-node-windows-amd64.tar.gz"
+        },
+        "ControlPlane" : {
+            "IpAddress" : "kubemasterIP",
+            "Username" : "localadmin",
+            "KubeadmToken" : "token",
+            "KubeadmCAHash" : "discovery-token-ca-cert-hash"
+        },
+        "KubeProxy" : {
```
daschott: For consistency, should we also add support for Kubelet gates? It looks like there is already some code in place to forward the args to kubelet

kubeadm/v1.15.0/KubeClusterHelper.psm1
```diff
@@ -0,0 +1,1360 @@
...
+    $env:path += ";$env:ProgramFiles\Docker"
+    # For persistent use after a reboot
+    $existingMachinePath = [Environment]::GetEnvironmentVariable("Path",[System.EnvironmentVariableTarget]::Machine)
+    [Environment]::SetEnvironmentVariable("Path", $existingMachinePath + ";$env:ProgramFiles\Docker", [EnvironmentVariableTarget]::Machine)
+
+    $cmd = get-command docker.exe -ErrorAction SilentlyContinue
+    if (!$cmd)
+    {
+        $dockerVersion = $Version
+        DownloadFile  "https://master.dockerproject.org/windows/x86_64/${dockerVersion}.zip" -Destination "$env:TEMP\$dockerVersion.zip"
```
daschott: This Docker version is very recent which is nice, but it fails the kubeadm pre-flight check for me :(

kubeadm/v1.15.0/KubeClusterHelper.psm1
```diff
@@ -0,0 +1,1360 @@
...
+    {
+        $Release = $Source.Release
+    }
+    $Url = "https://dl.k8s.io/v${Release}/kubernetes-node-windows-amd64.tar.gz"
+    if ($Source.Url)
+    {
+        $Url = $Source.Url
+    }
+    DownloadFile -Url  $Url -Destination $Global:BaseDir/kubernetes-node-windows-amd64.tar.gz
+    tar -zxvf $Global:BaseDir/kubernetes-node-windows-amd64.tar.gz -C $DestinationPath
```
daschott: *open question:* should we be cleaning up the tarball so it doesn't take up space on the machine/device?

kubeadm/v1.15.0/KubeCluster.ps1
ksubrmnn: Resolved. Missed a change.

kubeadm/v1.15.0/KubeClusterHelper.psm1
ksubrmnn: This gets cleaned up when we delete C:/ProgramData/Kubernetes (for me anyway)

kubeadm/v1.15.0/KubeCluster.ps1
ksubrmnn: This was an artifact from an old script. Removed

kubeadm/v1.15.0/KubeClusterHelper.psm1
ksubrmnn: Fixed by just checking docker images again

kubeadm/v1.15.0/Kubeclustervxlan.json
ksubrmnn: Let's add this in a future PR so this one doesn't get held up

kubeadm/v1.15.0/KubeCluster.ps1
ksubrmnn: Let's add this in a future PR so this one doesn't get held up

kubeadm/v1.15.0/KubeClusterHelper.psm1
ksubrmnn: Fixed by just checking docker images again

kubeadm/v1.15.0/KubeClusterHelper.psm1
ksubrmnn: ```Install-Module -Name DockerMsftProvider -Repository PSGallery -Force```

```Install-Package -Name docker -ProviderName DockerMsftProvider -Force -RequiredVersion 18.09  ```

Can you try updating the code with the above and see if that works?

**Comment by ksubrmnn** [comment]
@daschott I updated so now we pull cni binaries from containernetworkingplugins. Can you give this another try when you get a chance?

File: kubeadm/v1.15.0/KubeClusterHelper.psm1 [hunk 41]
```diff
@@ -0,0 +1,1353 @@
...
+
+function GetKubeDnsServiceIp()
+{
+    $svc = ConvertFrom-Json $(kubectl.exe get services -n kube-system -o json | Out-String)
+    $svc.Items | foreach { $i = $_; if ($i.Metadata.Name -match "dns") { return $i.spec.ClusterIP } }
+}
+
+function GetAPIServerEndpoint() {
+    $endpoints = ConvertFrom-Json $(kubectl.exe get endpoints --all-namespaces -o json | Out-String)
+    $endpoints.Items | foreach { $i = $_; if ($i.Metadata.Name -match "kubernetes") { return "$($i.subsets.addresses.ip):$($i.subsets.ports.port)" } }
```
bnason: I had trouble using this with kubernetes-dashboard being deployed in my cluster. Below was how I fixed it locally.

```suggestion
    $endpoints.Items | foreach { $i = $_; if ($i.Metadata.Name -eq "kubernetes") { return "$($i.subsets.addresses.ip):$($i.subsets.ports.port)" } }
```

kubeadm/v1.15.0/KubeClusterHelper.psm1
daschott: yes that works. I uninstalled Docker & re-validated to make sure.

daschott:
@ksubrmnn Thank you, LGTM. Additional ideas for a future PR worth considering could be:
 - Add support for kubelet feature gates
 - Add support for pinning down CNI plugin version to download

ksubrmnn:
@PatrickLang

PR #4, opened 2019-07-29
Fix wincat to correctly detect when to interrupt/disconnect

benmoss:
Previously wasn't closing if the server disconnected, which was causing
e2e tests to fail.

https://github.com/kubernetes/kubernetes/issues/78857

@PatrickLang

PR #6, opened 2019-08-13
bug: pause image shouldn't emit logs

PatrickLang:
Fixes https://github.com/kubernetes/kubernetes/issues/81298

(also includes #4 for testing)


PatrickLang:
@yujuhong - can you give this a look?

yujuhong:
Changing to pauseloop lgtm. Thanks!

benmoss:
if you wanted to go even further and reduce the memory footprnit you could use the pause.c code from https://github.com/kubernetes/kubernetes/pull/75618/files

PR #7, opened 2019-08-23
Switch to pauseloop with no logs

PatrickLang:
Replaces #6, same code. I mistakenly made that PR in a branch here instead of in my fork. That prevented me from cleaning things up with a rebase.

Fixes kubernetes/kubernetes#81298

Later once we have a Windows build solution in place (proposed in https://github.com/kubernetes/kubernetes/pull/75618/files), we should move wincat and pause builds into the k/k repo.


PatrickLang:
@benmoss @yujuhong @ddebroy - can one of you review?

PR #10, opened 2019-09-11
Linux node selector adder

KnicKnic:
see if the PR works now that I have moved code to third_party


benmoss:
Can we clarify what the purpose of this is? My understanding is it's something around automatically adding taints/tolerations/nodeSelectors to pods, but it seems like there's some shifting ideas around what the intended goal is now.

A standard cluster with Windows nodes should have a taint on those nodes that means that Windows pods need to opt-in to being scheduled on them. The flip-side of this is that Linux pods don't need to change at all in order to run, since they will not have the toleration and therefore ignore the Windows nodes. The burden is all on Windows devs to know they need to add the magic toleration and nodeSelectors to all their workoads.

It sounds from the work done here like you were running it in a cluster without any taints. The proposal then being that we can drop taints as a requirement but maintain the requirement that Windows workloads need `nodeSelector` requirements. This could be viable, but we'd need to get consensus around it. My main complaint about it is that compared to today's standard approach of taints, I'm not sure it significantly reduces the burden, you still need to ensure you have a nodeSelector on your Windows workloads.

KnicKnic:
> Can we clarify what the purpose of this is? My understanding is it's something around automatically adding taints/tolerations/nodeSelectors to pods, but it seems like there's some shifting ideas around what the intended goal is now.
> A standard cluster with Windows nodes should have a taint on those nodes that means that Windows pods need to opt-in to being scheduled on them. The flip-side of this is that Linux pods don't need to change at all in order to run, since they will not have the toleration and therefore ignore the Windows nodes. The burden is all on Windows devs to know they need to add the magic toleration and nodeSelectors to all their workoads.
> It sounds from the work done here like you were running it in a cluster without any taints. The proposal then being that we can drop taints as a requirement but maintain the requirement that Windows workloads need nodeSelector requirements. This could be viable, but we'd need to get consensus around it. My main complaint about it is that compared to today's standard approach of taints, I'm not sure it significantly reduces the burden, you still need to ensure you have a nodeSelector on your Windows workloads.

Originally I didn't remember seeing documentation for taints and tolerations called out on Kubernetes website specifically for windows. I see that it is now called out. https://kubernetes.io/docs/setup/production-environment/windows/user-guide-windows-containers/#taints-and-tolerations

However it does not seem that we have standardized on the specific taints  / best practices for windows. There is a suggestion to use runtime classes to help solve this problem. https://github.com/kubernetes/enhancements/blob/f1a799d5f4658ed29797c1fb9ceb7a4d0f538e93/keps/sig-node/runtime-class-scheduling.md.

As this is an evolving space I asked if this code would be useful as a discussion piece or solve the fact that there is currently no standard. This code could be flipped and add the right toleration for your distro of Kubernetes for the nodes based on the windows os selector.

If you feel that the taints have been standardized, than this is not needed except to support people that wrote code prior to standardization. A quick search on GitHub yields a bunch of people without tolerations on windows manifests, however they do specify os selector for windows.

benmoss:
Yeah, it's still a loose standard and definitely still evolving. I was surprised to discover yesterday that Rancher's product for instance seems to taint the Linux node and puts no taints on Windows nodes when you spin up a Windows cluster.

I guess I'd just like some docs talking about what problem this tool solves :cat:

KnicKnic:
> Yeah, it's still a loose standard and definitely still evolving. I was surprised to discover yesterday that Rancher's product for instance seems to taint the Linux node and puts no taints on Windows nodes when you spin up a Windows cluster.
> 
> I guess I'd just like some docs talking about what problem this tool solves 🐱. Doesn't need to be fancy or nothing.

I figured this was covered in the existing readme. If the text is not clear I'll take in any PR. :-)

PR #15, opened 2019-09-27
Use tilde (~) to resolve the home directory (vagrant case)

a4099181:
The environment variables `$env:HOMEDRIVE` and `$env:HOMEPATH` are empty
when vagrant executes `KubeCluster.ps1 -install`.
It makes _SSH key_ is not found even when it is correctly provisioned.
Script asks to generate _SSH key_ but it shouldn't do it.

### DIFF
--- a/kubeadm/v1.15.0/KubeCluster.ps1
+++ b/kubeadm/v1.15.0/KubeCluster.ps1
@@ -193,7 +193,7 @@ if ($install.IsPresent)
     InstallKubernetesBinaries -Destination  $Global:BaseDir -Source $Global:ClusterConfiguration.Kubernetes.Source
     DownloadCniBinaries -NetworkMode $Global:NetworkMode -CniPath $(GetCniPath)
 
-    if (!(Test-Path ~/.ssh/id_rsa.pub))
+    if (!(Test-Path "$(GetUserDir)/.ssh/id_rsa.pub"))
     {
         $res = Read-Host "Do you wish to generate a SSH Key & Add it to the Linux control-plane node [Y/n] - Default [Y]"
         if ($res -eq '' -or $res -eq 'Y'  -or $res -eq 'y')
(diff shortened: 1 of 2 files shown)


neolit123:
> The environment variables $env:HOMEDRIVE and $env:HOMEPATH are empty

i'm pretty sure you can set these env. variables in vagrant?

kubeadm/v1.15.0/KubeCluster.ps1
```diff
@@ -193,7 +193,7 @@ if ($install.IsPresent)
     InstallKubernetesBinaries -Destination  $Global:BaseDir -Source $Global:ClusterConfiguration.Kubernetes.Source
     DownloadCniBinaries -NetworkMode $Global:NetworkMode -CniPath $(GetCniPath)
 
-    if (!(Test-Path $env:HOMEDRIVE/$env:HOMEPATH/.ssh/id_rsa.pub))
+    if (!(Test-Path ~/.ssh/id_rsa.pub))
```
neolit123: does `~` work in PowerShell?

neolit123:
>  GitHub didn't allow me to assign the following users: ksubrmnn.

@ksubrmnn given you are already a member of the kubernetes org, you are eligible for kubernetes-sigs too. if you file an  issue here https://github.com/kubernetes/org/issues you can become a member.

kubeadm/v1.15.0/KubeCluster.ps1
neolit123: EDIT: looks like it does.

a4099181: Yes, it works well. I tested it.

a4099181:
> > The environment variables $env:HOMEDRIVE and $env:HOMEPATH are empty
> 
> i'm pretty sure you can set these env. variables in vagrant?

Yes, I can. But in my opinion, this is not the best option to fix it in Vagrantfile. The key problem of that case will not disappear and still, sometimes the script will stop automation.

a4099181:
I think _Active Directory_ is an important case because `$env:HOMEDRIVE` and `$env:HOMEPATH` may point to a remote user profile when a user is logged into a domain (AD).
In this case `~` and `$env:USERPROFILE` still points to a local user profile.
If so, `~` may be used as a fallback for the case when `$env:HOMEDRIVE` and `$env:HOMEPATH` are empty.
This is my alternative solution implemented in 2nd commit.

ksubrmnn:
@neolit123 will take a look at this next week. And I'll join Kubernetes-sigs as well

ksubrmnn:
lgtm

PR #16, opened 2019-09-28
Fix windows support on 1903, add v1.16.0 support, add unattended mode/general cleanup

superlime:
Logic broken up into multiple commits:
- Support windows 1903 LWVNIC
  - Adds support for new LWVNICs used in 1903 (without breaking backwards compatibility)
- Add configs for v1.16.0, move non-version-specific files to base kubeadm dir
  - Both script files moved to /kubeadm
  - Added new dir for v1.16.0 with configs supporting that version
- Add fully unattended mode to scripts, general fit/finish fixes
  - Add new flag `-force` that enables fully unattended script execution
  - Populates ~/.ssh/known_hosts with key from master specified in kubecluster.config if in unattended mode
  - Move help output to use PowerShell comment-based-help (also enables standard powershell get-help)
  - Fix user home paths to sue $env:USERPROFILE instead of $env:HOMEDRIVE/$env:HOMEPATH, as those aren't available in some session types/windows skus.  $env:USERPROFILE appears to be always available, though.
  - Fix reset mode's folder deletions to include the -Recurse flag
- Update CNI config to pass proper DNS configuration
  - Fix CNI configuration to use updated dns settings
- Clean up scheduled task if rebooting in prereqs
  - Scheduled task was left behind previously, and could cause problems in the case of bad configs.  Properly removed once prereq install completes successfully now.
  - Properly passes through and respects -force flag for unattended setup even if rebooting
- Fix for ClusterDNS warning
  - Uncommented kubelet config lines for adding clusterdns configuration
  - Added flags to kubelet service definition for clusterdns configuration
  - If these were missing, sometimes you'd get dns issues or warnings about ClusterDNS not being set in pod details

Validated with windows nodes running 1809 and 1903.


superlime:
I signed it

superlime:
@ksubrmnn

ksubrmnn:
lgtm

superlime:
@PatrickLang

kubeadm/v1.16.0/Kubeclustervxlan.json
```diff
@@ -0,0 +1,44 @@
+{
+    "Cri" : {
+        "Name" : "dockerd",
+        "Images" : {
+            "Pause" : "mcr.microsoft.com/k8s/core/pause:1.2.0",
+            "Nanoserver" : "mcr.microsoft.com/windows/nanoserver:1809",
```
ddebroy: Do these need to be adjusted for Windows 1903?

kubeadm/v1.16.0/KubeCluster.ps1
```diff
@@ -0,0 +1,364 @@
+<#
```
ddebroy: Will it make sense to refactor the structure here to share code between the different k8s versions? Looks like a lot of the 1.16 additions share common code with 1.15.

superlime: > Will it make sense to refactor the structure here to share code between the different k8s versions? Looks like a lot of the 1.16 additions share common code with 1.15.

I was thinking about that when I was making the checkin..  The diff between 1.15 and 1.16 at this point is really just settings in the config files.  There's defaults in the scripts that change, but that's overridden by the config files.  At some point in the future there might be a wider diff in flags/etc needed though...

My initial thought was to just remove the defaults from the scripts entirely and have them be version agnostic (with additional errors if values weren't passed in in the config files), but wasn't sure if that'd make things too complicated down the road if there needed to be actual script differences between versions.

> I was not sure which changes in this PR are specifically needed for 1903.

Ahh sorry, should have been more clear in the PR notes there.  The 1903-specific support is lines 451/452 of KubeClusterHelper.psm1.  In 1903+ apparently the NICs used by HNS switch to a new LWVNIC model that doesn't show up in Win32_NetworkAdapterConfiguration, so need to be discovered via this registry node.  I verified that the new method still works for older windows versions (so not breaking backwards compatibility).

kubeadm/v1.16.0/Kubeclustervxlan.json
superlime: Yeah, but I think the default probably should stay at 1809?  The main changes just allow the script to work with any released version of windows as a worker node, but I think the default probably makes sense to stay at 1809.  Probably there should be documentation changes to tell people to use the right images for the right windows node OS version?

Alternately we could provide config files tailored for each OS release/kubernetes release, but that seems like it'd quickly get out of hand...

superlime: Or another idea, we could remove the values from the config file for the default case and automatically pick the right images based on the host's OS version.  Still would be useful to have this as a configuration option for the user to be able to override the default versions (i.e. running against custom images or unsupported os versions), though.

ddebroy: I see. I think the default values pointing to images that work on the LTS release is fine for now. Advanced/SAC users can override to the SAC versions as needed.

kubeadm/v1.16.0/KubeCluster.ps1
superlime: I'll reorganize so the scripts are in the root level of the repo and just the configs are in the version-specific directories (since they're all that's actually version-specific at this point).

pjh:
> Support windows 1903 LWVNIC

What is LWVNIC?

Thanks for breaking the PR up into separate commits - the changes required for 1903 are crucial for us.

ddebroy:
Thanks for the changes, @superlime 
/lgtm

kubeadm/KubeCluster.ps1
```diff
@@ -308,7 +358,7 @@ elseif ($Reset.IsPresent)
     UninstallKubelet
     UninstallKubernetesBinaries -Destination  $Global:BaseDir
```
daschott: Can we also clean up the registered task here?
`Unregister-ScheduledTask -TaskName "containerbootstrap"`

If this isn't done then users who had an  erroneous configuration once will not be able to overwrite the stale task upon subsequent installation.

superlime: Working on check/fix for this.  Should have an additional commit by EOD.

superlime:
Rebased, added requested fix for cleaning up the scheduled task, fixed additional spots that needed -force or -recurse, uncommented lines for adding clusterdns configuration (according to @daschott & @ksubrmnn they're needed) and added flags to the service definition.

Should be good to go, finally.  :)

daschott:
LGTM

PR #19, opened 2019-10-10
Fixed GetAPIServerEndpoint if there is one master

### LINKED ISSUE #18: KubeCluster 1.15 Scripts fail to join with one master
After PR #12 the script fails to join a cluster with a single master.

This is because in GetAPIServerEndpoint, $i.subsets.addresses.ip is not an array if there is one master. Because of this, the API server endpoint is returned as 1:6443 where 1 represents the first letter of the IP address. So we get the first character of the IP instead of the first IP.

a4099181:
Refers to issue #18

### DIFF
--- a/kubeadm/v1.15.0/KubeClusterHelper.psm1
+++ b/kubeadm/v1.15.0/KubeClusterHelper.psm1
@@ -1207,7 +1207,7 @@ function GetKubeDnsServiceIp()
 
 function GetAPIServerEndpoint() {
     $endpoints = ConvertFrom-Json $(kubectl.exe get endpoints --all-namespaces -o json | Out-String)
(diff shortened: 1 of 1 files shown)


ddebroy:
Thanks for the PR! Can you please update the description of the PR with an actual link to issue 18?

a4099181:
@ddebroy OK, done. Now I also know, that I cann't use a keyword like _fixes_ in a commit message.

/assign @patricklang

superlime:
I'd fixed single-master in #16 (not realizing the multi-master fix was in).  I'll update the other PR to remove those changes once this goes through.  :)

ddebroy:
cc @ksubrmnn to lgtm

ksubrmnn:
lgtm

PR #13, opened 2019-09-25
Wait for management IP interface before FlannelD service will start.

a4099181:
FlannelD attempts to start the service immediately after creating
an external network. It's so fast that the network interface
is not available on time.

FlannelD service start fails with the following output:
```
I(...)    6124 main.go:450] Searching for interface using <ManagementIP>
I(...)    6124 main.go:210] Could not find valid interface matching <ManagementIP>: error looking up interface <ManagementIP>: Interface not found: <ManagementIP>
E(...)    6124 main.go:234] Failed to find interface to use that matches the interfaces and/or regexes provided
```

PR #30, opened 2019-11-26
Fixed hang/fail on FlannelD create of VXLAN

kellybirr:
This change fixes an issue where 'FlannelD' service sometimes fails to create 'vxlan0' on first start. This change makes one attempt to turn it off and back on again before giving up and timing out.

This fixed the issue for me completely on Windows Server 1909 (SAC) with Kubernetes 1.16.3

### DIFF
--- a/kubeadm/KubeCluster.ps1
+++ b/kubeadm/KubeCluster.ps1
@@ -342,8 +342,16 @@ if ($Join.IsPresent)
     if ($Global:Cni -eq "flannel")
     {
         CreateExternalNetwork -NetworkMode $Global:NetworkMode -InterfaceName $Global:InterfaceName
+        WaitForNetworkInterface -IpAddress $Global:ManagementIp
         StartFlanneld 
-        WaitForNetwork $Global:NetworkName
+        
+        # Ensure Flannel created the HNS network
+        $flannelNetOk = WaitForNetwork -NetworkName $Global:NetworkName -DieOnFail $false
+        if (!$flannelNetOk)
+        {
+            RestartFlanneld
+            WaitForNetwork -NetworkName $Global:NetworkName -DieOnFail $true
+        }
     }
 
     # 3. Install & Start Kubeproxy
--- a/kubeadm/KubeClusterHelper.psm1
+++ b/kubeadm/KubeClusterHelper.psm1
@@ -240,6 +240,7 @@ function CleanupPolicyList()
 function WaitForNetwork($NetworkName, $waitTimeSeconds = 60)
 {
     $startTime = Get-Date
+    $triedRestart = $false
 
     # Wait till the network is available
     while ($true)
@@ -247,7 +248,21 @@ function WaitForNetwork($NetworkName, $waitTimeSeconds = 60)
         $timeElapsed = $(Get-Date) - $startTime
         if ($($timeElapsed).TotalSeconds -ge $waitTimeSeconds)
         {
-            throw "Fail to create the network[($NetworkName)] in $waitTimeSeconds seconds"
+            if ($triedRestart)
+            {
+                throw "Fail to create the network[($NetworkName)] in $waitTimeSeconds seconds, again."
+            }
+            
+            # Try Restarting Flanneld and see if that fixed the problem
+            Write-Host "Network [($NetworkName)] was not created within $waitTimeSeconds seconds..."
+            Write-Host "Attempting to turn FlannelD service off and back on again."
+            Stop-Service -Name FlannelD
+            Start-Sleep 5
+            StartFlanneld
+
+            # restart countdown & set triedRestart flag
+            $startTime = Get-Date
+            $triedRestart = $true
         }
         if ((Get-HnsNetwork | ? Name -EQ $NetworkName.ToLower()))
         {
--- a/kubeadm/KubeClusterHelper.psm1
+++ b/kubeadm/KubeClusterHelper.psm1
@@ -261,7 +261,7 @@ function WaitForNetwork($NetworkName, $WaitTimeSeconds = 60, $DieOnFail = $true)
         }
         if ((Get-Service -Name FlannelD).Status -ne "Running")
         {
-            Write-Host "FlannelD service stopped unexpectecdly"
+            Write-Host "FlannelD service stopped unexpectedly"
             return $false
         }
         Write-Host "Waiting for the Network ($NetworkName) to be created by flanneld"
--- a/kubeadm/KubeClusterHelper.psm1
+++ b/kubeadm/KubeClusterHelper.psm1
@@ -237,39 +237,63 @@ function CleanupPolicyList()
     }
 }
 
-function WaitForNetwork($NetworkName, $waitTimeSeconds = 60)
+function WaitForNetwork($NetworkName, $WaitTimeSeconds = 60, $DieOnFail = $true)
 {
     $startTime = Get-Date
-    $triedRestart = $false
 
     # Wait till the network is available
     while ($true)
     {
         $timeElapsed = $(Get-Date) - $startTime
-        if ($($timeElapsed).TotalSeconds -ge $waitTimeSeconds)
+        if ($($timeElapsed).TotalSeconds -ge $WaitTimeSeconds)
         {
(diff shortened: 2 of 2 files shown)


neolit123:
please sign the CLA as the bot mentions above.

kellybirr:
thanks, just signed

kubeadm/KubeClusterHelper.psm1
```diff
@@ -240,14 +240,29 @@ function CleanupPolicyList()
...
+            {
+                throw "Fail to create the network[($NetworkName)] in $waitTimeSeconds seconds, again."
+            }
+            
+            # Try Restarting Flanneld and see if that fixed the problem
+            Write-Host "Network [($NetworkName)] was not created within $waitTimeSeconds seconds..."
+            Write-Host "Attempting to turn FlannelD service off and back on again."
+            Stop-Service -Name FlannelD
+            Start-Sleep 5
+            StartFlanneld
```
neolit123: > service sometimes fails to create 'vxlan0' on first start

this is odd. can you provide more details on why it happens exactly?

neolit123: random though: would waiting for N seconds before the flannel service is started also solve this problem?

kellybirr: I have a Kubernetes cluster built of VMs that I've been testing on, running Kubernetes 1.16.3.  The  sole master and 5 workers run Ubuntu 18.04.3 and for the last few day's i've been working on add Windows nodes.  I have 4 windows nodes, currently running sac-1909, but I've observed the exact same behavior on ltsb-2019.  

When FlannelD starts, the first time, a vxlan0 HSN network does not get created and the ps script times out waiting.  In every case, on every test VM, simply restarting FlannelD solves this problem.  I've run this altered script on 4 nodes, today, and on each node it timed out, the 1st time,then restarted and succeeded in the -join operation.

I suspect this is actually a FlannelD issue, but this script change works around the issue and allows the join to succeed and the node to work properly.

kellybirr: As to waiting longer; that was actually my 1st thought.  I tried going as high as 300 seconds and that did not solve the issue.  I suspect that FlannelD is actually silently failing on the 1st start, but succeeding after that.  In all of my tests, starting FlannelD again was the only solution for me.

kellybirr: Ultimately, I realize this is a work-around at best and would probably be called a "hack" by most. A real fix to FlannelD or Windows would be a better solution.  

However, work-arounds are something that we Windows admins are all to familiar with.

neolit123: are there any logs indicating what is the failure?

neolit123: > Ultimately, I realize this is a work-around at best and would probably be called a "hack" by most. A real fix to FlannelD or Windows would be a better solution.

if we can point the flannel maintainers to the the exact failure this can be solved on their side.

kellybirr:
I just looked at PR#13 (https://github.com/kubernetes-sigs/sig-windows-tools/pull/13).  I think it's possible my issue may be related.  I'm going to tear down and re-build one of my nodes and try that branch to see if it resolves my issue.  If so, that fix is better IMO.

neolit123:
> I just looked at PR#13 (#13). I think it's possible my issue may be related. I'm going to tear down and re-build one of my nodes and try that branch to see if it resolves my issue. If so, that fix is better IMO

it does look related.

kellybirr:
It may be related, however i tried the code changes in PR #13, exactly as done there, but against the newest 'master' versions.  It resulted in the exact same problem.  My "external" interface is created properly, but the 'vxlan0' network is not created until FlannelD is started a 2nd time.  This is on a brand new VM build of SAC-1909.

kellybirr:
Attached here are the output of ".\KubeCluster.ps1 -ConfigFile .\Kubeclustervxlan.json -join" and my "flanneldsvc.log".

[kubecluster-join-output.txt](https://github.com/kubernetes-sigs/sig-windows-tools/files/3895149/kubecluster-join-output.txt)

[flanneldsvc.log](https://github.com/kubernetes-sigs/sig-windows-tools/files/3895148/flanneldsvc.log)

kellybirr:
I ran the original sources again today and it still looks a lot like #13 in these logs.  I'm going to keep trying to figure out why the changes from that PR don't fix it for me.

[flanneldsvc.log](https://github.com/kubernetes-sigs/sig-windows-tools/files/3897501/flanneldsvc.log)

kellybirr:
OK, my issue is definitely related to #13! However, the changes in that PR don't work as-is for me.  

My VMs need to sleep at least once, if the interface is found on the first pass in `WaitForNetworkInterface`. It appears that even though `Get-NetIPAddress` can find it right away, FlannelD may not without at least a 5 second sleep after `CreateExternalNetwork` is called. 

I have a modified version of the changes from PR-13, based on the most up-to-date master scripts for v1.16.  

What is the proper protocol here?  I'm not sure proposing an amendment to that PR is the best course as it's based on older versions of the scripts.  At the same time, it feels rude to mostly duplicate someone else' code in a new PR?

neolit123:
thanks for your investigation @kellybirr 
@PatrickLang @ksubrmnn do you know why this is happening and how to proceed with fixing it?

kellybirr:
I'm on the slack channel for sig-windows-tools, i can reliably reproduce the problem and have a modified version of #13 that does resolve the issue for me.  I'm happy to have a slack call or screen share, if that would help? 

I have a solution to my issue, and i can move forward.  I'm just hoping to share my solution and save others' from losing 2-3 days troubleshooting a timing issue, like i have.

neolit123:
@kellybirr if you have a better version of #13 just send the PR?
ideally i want review and feedback from Flannel savvy folks.

kellybirr:
OK, I'll update the code in this PR to reflect my latest and best changes, later today

kellybirr:
I just posted my latest changes, that incorporate the work of PR #13, based on the newest master scripts and also incorporate the possible need to stop and/or restart FlannelD, but in a much cleaner way than my original code.

I've tested this, multiple times, through both possible paths on 2 VMs running SAC-1909.

kellybirr:
I can also test this on a VM with LTSB-2019 over the long weekend, if that helps.

kubeadm/KubeCluster.ps1
```diff
@@ -342,8 +342,16 @@ if ($Join.IsPresent)
...
         CreateExternalNetwork -NetworkMode $Global:NetworkMode -InterfaceName $Global:InterfaceName
+        WaitForNetworkInterface -IpAddress $Global:ManagementIp
         StartFlanneld 
-        WaitForNetwork $Global:NetworkName
+        
+        # Ensure Flannel created the HNS network
+        $flannelNetOk = WaitForNetwork -NetworkName $Global:NetworkName -DieOnFail $false
+        if (!$flannelNetOk)
+        {
+            RestartFlanneld
```
neolit123: yet, we are still forcing a flannel restart.
at this point and ideally, i want the assigned reviewers to take a look at this.

kellybirr: makes sense... we can also cut that bit out.  In my latest tests the restart does not trigger, `WaitForNetworkInterface` takes care of the issue.  I just left this capability to restart FlannelD, as a last resort.

ksubrmnn:
@JocelynBerrendonner

JocelynBerrendonner:
> @JocelynBerrendonner

@dineshgovindasamy

michmike:
@benmoss  your thoughts on this PR?

kellybirr:
I'm around if anyone has questions.

benmoss:
This is for the old set of scripts which we are planning to remove

neolit123:
hi, @kellybirr 
we are deprecating the old scripts (Aalpha) and now the feature is graduating to Beta with a new set of scripts:
https://github.com/kubernetes-sigs/sig-windows-tools/tree/master/kubeadm

thanks
/close

kellybirr:
Thanks for the heads up

PR #34, opened 2020-02-24
Add scripts/manifests for new daemonset kubeadm approach

benmoss:
@neolit123 @PatrickLang @michmike 

- Moved all the images to https://hub.docker.com/orgs/sigwindowstools which I can grant others access to
- Using nssm to allow for dynamic kubelet args via the `kubeadm-flags.env` file

Finalizing documentation here https://github.com/kubernetes/website/pull/19217


kubeadm/flannel/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+RUN mkdir C:\cni; \
+  pushd C:\cni; \
+  curl.exe -Lo cni.tgz https://github.com/containernetworking/plugins/releases/download/v${env:cniVersion}/cni-plugins-windows-amd64-v${env:cniVersion}.tgz; \
+  tar -xf cni.tgz; \
+  rm cni.tgz
+
+RUN curl.exe -Lo C:\windows\wins.exe https://github.com/rancher/wins/releases/download/v0.0.4/wins.exe
+
+RUN curl.exe -Lo C:\windows\yq.exe https://github.com/mikefarah/yq/releases/download/2.4.1/yq_windows_amd64.exe
```
neolit123: i don't think we should install wins.exe and yq.exe to c:\windows. the system root is for windows originated software.

kubeadm/flannel/flannel-host-gw.yml
```diff
@@ -0,0 +1,171 @@
...
+    cp -force /etc/kube-flannel/net-conf.json /host/etc/kube-flannel
+    cp -force -recurse /cni/* /host/opt/cni/bin
+    cp -force /flannel/flanneld.exe /host/flannel/flanneld.exe
+    cp -force /kube-proxy/kubeconfig.conf /host/flannel/kubeconfig.yml
+    cp -force /var/run/secrets/kubernetes.io/serviceaccount/* /host/flannel/var/run/secrets/kubernetes.io/serviceaccount/
+    wins cli process run --path C:\flannel\flanneld.exe --args "--kube-subnet-mgr --kubeconfig-file /flannel/kubeconfig.yml" --envs "POD_NAME=$env:POD_NAME POD_NAMESPACE=$env:POD_NAMESPACE"
+  cni-conf.json: |
+    {
+      "name": "cbr0",
+      "cniVersion": "0.3.0",
```
neolit123: the `cniVersion` in `kubeadm/flannel/Dockerfile` is 0.8.5.
can we have 0.8.5 here as well?

kubeadm/flannel/flannel-overlay.yml
```diff
@@ -0,0 +1,160 @@
...
+    cp -force /etc/kube-flannel/net-conf.json /host/etc/kube-flannel
+    cp -force -recurse /cni/* /host/opt/cni/bin
+    cp -force /flannel/flanneld.exe /host/flannel/flanneld.exe
+    cp -force /kube-proxy/kubeconfig.conf /host/flannel/kubeconfig.yml
+    cp -force /var/run/secrets/kubernetes.io/serviceaccount/* /host/flannel/var/run/secrets/kubernetes.io/serviceaccount/
+    wins cli process run --path C:\flannel\flanneld.exe --args "--kube-subnet-mgr --kubeconfig-file /flannel/kubeconfig.yml" --envs "POD_NAME=$env:POD_NAME POD_NAMESPACE=$env:POD_NAMESPACE"
+  cni-conf.json: |
+    {
+      "name": "flannel.4096",
+      "cniVersion": "0.3.0",
```
neolit123: [1]

kubeadm/kube-proxy/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+ARG k8sVersion
+
+RUN mkdir C:\kube-proxy; \
+  pushd C:\kube-proxy; \
+  curl.exe -LO https://dl.k8s.io/v${env:k8sVersion}/bin/windows/amd64/kube-proxy.exe
+
+RUN curl.exe -Lo C:\windows\wins.exe https://github.com/rancher/wins/releases/download/v0.0.4/wins.exe
+
+RUN curl.exe -Lo C:\windows\yq.exe https://github.com/mikefarah/yq/releases/download/2.4.1/yq_windows_amd64.exe
```
neolit123: [2]

kubeadm/kube-proxy/kube-proxy.yml
```diff
@@ -0,0 +1,94 @@
...
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              apiVersion: v1
+              fieldPath: spec.nodeName
+        - name: POD_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
+        image: sigwindowstools/kube-proxy:1.17.3
```
neolit123: ideally we should not pin the kube proxy version.
is there a way for the user to customize this during the initial script execution - e.g. using search and replace (`sed`)?

kube-proxy has a skew of +1 / -1 releases compared to core kube, so if this version is not updated and someone tries to deploy e.g. 1.19 it might fail.

kubeadm/kube-proxy/Dockerfile
```diff
@@ -0,0 +1,15 @@
+ARG k8sVersion="1.17.3"
```
neolit123: [3]

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -0,0 +1,83 @@
...
+
+mkdir -force C:\var\log\kubelet
+mkdir -force C:\var\lib\kubelet\etc\kubernetes
+mkdir -force C:\etc\kubernetes\pki
+New-Item -path C:\var\lib\kubelet\etc\kubernetes\pki -type SymbolicLink -value C:\etc\kubernetes\pki\
+
+$StartKubeletFileContent = '$FileContent = Get-Content -Path "/var/lib/kubelet/kubeadm-flags.env"
+$global:KubeletArgs = $FileContent.Trim("KUBELET_KUBEADM_ARGS=`"")
+
+$cmd = "C:\k\kubelet.exe $global:KubeletArgs --cert-dir=$env:SYSTEMDRIVE\var\lib\kubelet\pki --config=/var/lib/kubelet/config.yaml --bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf --kubeconfig=/etc/kubernetes/kubelet.conf --hostname-override=$(hostname) --pod-infra-container-image=`"mcr.microsoft.com/k8s/core/pause:1.2.0`" --enable-debugging-handlers --cgroups-per-qos=false --enforce-node-allocatable=`"`" --network-plugin=cni --resolv-conf=`"`" --log-dir=/var/log/kubelet --logtostderr=false --image-pull-progress-deadline=20m"
```
neolit123: was wondering if we should put \kube-proxy and \flannel under c:\k as well.

neolit123:
thanks for adding this @benmoss @gab-satchi 
added some comments, mostly minor.

what is the plan for deprecating and removing the old code here?
https://github.com/kubernetes-sigs/sig-windows-tools/tree/master/kubeadm

kubeadm/flannel/flannel-host-gw.yml
benmoss: That's the version of the [CNI spec](https://github.com/containernetworking/cni/blob/master/SPEC.md), 0.8.5 is the version of the [CNI plugins bundle](https://github.com/containernetworking/plugins/releases/tag/v0.8.5)

kubeadm/flannel/Dockerfile
benmoss: Yeah, I was just being lazy since it's one of the few places that's on the PATH by default. I will update to somewhere else.

kubeadm/kube-proxy/kube-proxy.yml
benmoss: Yeah, I was not sure what to do about this problem. This manifest is not deployed/edited by a script at all, the user is doing a `kubectl apply` with this file directly. So we can document in the step where we direct the user to do this that they pick the version that matches their node/cluster. We can provide a sed command to do so, that's pretty straightforward.

kubeadm/kube-proxy/Dockerfile
benmoss: Yeah this is just the default, and so can be overridden with `docker build --build-arg k8sVersion="1.18.0"`. We'll need to set up some automation to produce new images on new releases though.

kubeadm/kube-proxy/kube-proxy.yml
neolit123: ideally it should be propagated from $KubernetesVersion in PrepareNode, so when the user picks a version the PowerShell patches the YAMLs that later the user applies.

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -0,0 +1,83 @@
...
+
+Param(
+    [parameter(Mandatory = $true, HelpMessage="Kubernetes version to use")]
+    [string] $KubernetesVersion
+)
+$ErrorActionPreference = 'Stop'
+
+function DownloadFile($destination, $source) {
+    Write-Host("Downloading $source to $destination")
+    curl.exe --fail -Lo $destination $source
```
neolit123: hm, i don't think curl is available on Windows by default. i remember having to download it when the last time i've tested.

kubeadm/kube-proxy/Dockerfile
neolit123: i've been playing with GitHub actions in this regard so automation seems possible.

but it this goes back to the topic of whether we want to branch / tag this repository and or have the install scripts in folders.

i think this is always the safest bet, because otherwise if a single set of scripts and docker files is maintained for the whole support skew of k8s it might end up with some `if version ...` branches.

kubeadm/scripts/PrepareNode.ps1
benmoss: :+1:

kubeadm/kube-proxy/kube-proxy.yml
benmoss: The problem is that `PrepareNode`/``InstallNssm` are run on the Windows nodes, whereas `kube-proxy.yml` is run from your workstation where you have `kubectl` access.

kubeadm/kube-proxy/Dockerfile
benmoss: What do you think of using github releases for the script and manifest artifacts? I don't want to create version subdirectories since they are going to be 99% identical to each other.

kubeadm/scripts/PrepareNode.ps1
benmoss: It's available by default from Windows 1803 on, so on every version that we're targeting

benmoss:
> what is the plan for deprecating and removing the old code here?
> https://github.com/kubernetes-sigs/sig-windows-tools/tree/master/kubeadm

Not sure, what do you think we should do? Delete it? Make some "archive" subdirectory or something?

neolit123:
> Not sure, what do you think we should do? Delete it? Make some "archive" subdirectory or something?

given:
- the feature state is currently alpha
- the old code is stored in git

i think we can just delete the old code in a new PR.

kubeadm/scripts/PrepareNode.ps1
neolit123: interesting, you are probably right.

kubeadm/kube-proxy/Dockerfile
neolit123: having releases / tags is fine, but without branches it will not allow backports.
one option is to start with a linear tag increments in a single branch (master), but if a change is required for a backport a branch can be created.

kubeadm/kube-proxy/kube-proxy.yml
neolit123: ok, i overlooked this. i guess it makes sense to document this in that case and give a `sed` example. my guess is that we would not want the users to apply the stock manifests without making sure it has the version they want.

benmoss:
I actually just realized that if we remove the old code it will break the old documentation. It links to things like https://github.com/kubernetes-sigs/sig-windows-tools/archive/master.zip that is just a zip of current master of the repo. Not sure if that is a concern or not.

neolit123:
> I actually just realized that if we remove the old code it will break the old documentation. It links to things like https://github.com/kubernetes-sigs/sig-windows-tools/archive/master.zip that is just a zip of current master of the repo. Not sure if that is a concern or not.

i'd leave it to sig-windows to establish the level of concern for this issue.
given this is an alpha feature the concerns might not be too high.

k/website have a no-backport policy, but a backport for a old dev-1.x branch there might be needed here. using HEAD for documentation links is a bad idea and ideally the old docs should be updated to point to commit SHAs.

neolit123:
thanks for the updates.
this seems fine to me as much as i can read powershell.

/lgtm
/hold
feel free to cancel the hold.

ksubrmnn:
@JocelynBerrendoner

ksubrmnn:
@JocelynBerrendonner

kubeadm/scripts/InstallNssm.ps1
```diff
@@ -0,0 +1,19 @@
+$ErrorActionPreference = "Stop"
+$global:NssmInstallDirectory = "$env:ProgramFiles\nssm"
+$arch = "win32"
+if ([Environment]::Is64BitOperatingSystem) {
+    $arch = "win64"
+}
+
+mkdir -Force $global:NssmInstallDirectory
+curl.exe -Lo nssm.zip https://nssm.cc/release/nssm-2.24.zip
+tar C $global:NssmInstallDirectory -xvf .\nssm.zip --strip-components 2 */$arch/*.exe
```
marosset: We had to mirror the nssm bits due to frequent flakes observed when pulling form https://nssm.cc

For Azure we re-hosted it at https://k8stestinfrabinaries.blob.core.windows.net/nssm-mirror/nssm-2.24.zip

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -0,0 +1,96 @@
...
+Write-Host "Registering wins service"
+wins.exe srv app run --register
+start-service rancher-wins
+
+mkdir -force C:\var\log\kubelet
+mkdir -force C:\var\lib\kubelet\etc\kubernetes
+mkdir -force C:\etc\kubernetes\pki
+New-Item -path C:\var\lib\kubelet\etc\kubernetes\pki -type SymbolicLink -value C:\etc\kubernetes\pki\
+
+$StartKubeletFileContent = '$FileContent = Get-Content -Path "/var/lib/kubelet/kubeadm-flags.env"
```
neolit123: just pointing out that this PR will improve the kubelet flag management for kubeadm on Windows:
https://github.com/kubernetes/kubernetes/pull/88287

PR #35, opened 2020-02-28
Move from deprecated beta node labels

benmoss:
A tiny change I just noticed 😸 . The beta ones are deprecated and scheduled for removal in 1.18, according to [this comment](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/well_known_labels.go#L30) anyway


benmoss:
@neolit123

PR #39, opened 2020-03-09
Add script for deploying/testing a cluster on GCE

benmoss:
Depends on windows-tools and k/k, assumes gcloud has already been configured. Probably needs another wrapper script to deal with Boskos

@neolit123

### DIFF
--- a/kubeadm/hack/e2e-cluster-gcp.sh
+++ b/kubeadm/hack/e2e-cluster-gcp.sh
@@ -1,0 +1,133 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+NODE_COUNT=${NODE_COUNT:-2}
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
+CLUSTER_NAME=${CLUSTER_NAME:-"test-$(date +%s)"}
+ARTIFACTS="${ARTIFACTS:-${PWD}/_artifacts}"
+scratchDir=$(mktemp -d -t $CLUSTER_NAME)
+export KUBECONFIG="$scratchDir/kubeconfig.yml"
+
+export ARTIFACTS
+mkdir -p "${ARTIFACTS}/logs"
+
+# disable gcloud prompts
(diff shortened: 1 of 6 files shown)


kubeadm/hack/e2e-cluster-gcp.sh
```diff
@@ -0,0 +1,139 @@
...
+kubectl -n kube-system wait --for=condition=Ready pods --all --timeout=15m
+
+# setting this env prevents ginkgo e2e from trying to run provider setup
+export KUBERNETES_CONFORMANCE_TEST="y"
+
+$(go env GOPATH)/src/sigs.k8s.io/windows-testing/gce/run-e2e.sh \
+	--provider=skeleton --num-nodes=$(expr $NODE_COUNT + 1) \
+	--ginkgo.focus="$FOCUS" --ginkgo.skip="$SKIP" \
+	--node-os-distro=windows --report-dir="$ARTIFACTS" \
+	--disable-log-dump=true
```
neolit123:
```
  # ginkgo regexes
  SKIP="${SKIP:-}"
  FOCUS="${FOCUS:-"\\[Conformance\\]"}"
  # if we set PARALLEL=true, skip serial tests set --ginkgo-parallel
  if [ "${PARALLEL:-false}" = "true" ]; then
    export GINKGO_PARALLEL=y
    if [ -z "${SKIP}" ]; then
      SKIP="\\[Serial\\]"
    else
      SKIP="\\[Serial\\]|${SKIP}"
    fi
  fi
```

neolit123: $NODE_COUNT should be N workers

neolit123: as mentioned on slack i think we should have something like that here above the `run-e2e.sh` call:
```
# split k8sVersion into major, minor
pushd $(go env GOPATH)/src/k8s.io/kubernetes/kubernetes
git fetch https://github.com/kubernetes/kubernetes.git release-$major.$minor
git checkout release-$major.$minor
popd
```

logs of what prow does:
https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-1-17/1237332976740601859/clone-log.txt

kubeadm/hack/startup/controlplane.sh
```diff
@@ -0,0 +1,70 @@
...
+apt-get update && apt-get install -y apt-transport-https curl
+curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
+cat <<EOF | tee /etc/apt/sources.list.d/kubernetes.list
+deb https://apt.kubernetes.io/ kubernetes-xenial main
+EOF
+apt-get update
+apt-get install -y kubelet kubeadm kubectl
+apt-mark hold kubelet kubeadm kubectl
+
+externalIP=$(curl -H "Metadata-Flavor:Google" 169.254.169.254/computeMetadata/v1/instance/network-interfaces/0/access-configs/0/external-ip)
```
neolit123: wondering what this line does and where is `169.254.169.254` coming from?
would this hardcoded IP work for the Boskos setup?

benmoss: This is the GCP metadata server, same as `metadata.google.internal` maybe that's clearer. Should work from any GCP VM.

neolit123: ok, got it.

PR #45, opened 2020-03-11
Add github action to publish kube-proxy images

benmoss:
You can see a successful build of this on my fork [here](https://github.com/benmoss/sig-windows-tools/actions/runs/53251162).

I set up a [bot account](https://hub.docker.com/u/sigwindowstoolsbot) on dockerhub and granted it write access to just the kube-proxy image.

One caveat is that this does seem to update the tag even when there are no changes in the base layers. I don't know exactly why or if this is going to be a problem.

/assign @neolit123


michmike:
i am approving but want @neolit123 to also take a look. (cancel the hold if you are good with this)
/hold

.github/workflows/kube-proxy.yml
```diff
@@ -0,0 +1,17 @@
...
+  build:
+    runs-on: windows-2019
+    steps:
+    - uses: actions/checkout@v2
+    - uses: azure/docker-login@v1
+      with:
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_SECRET }}
+    - name: Build
+      run: go run kubeadm/hack/publish-kubeproxy.go kubeadm/kube-proxy
```
neolit123: i don't think `windows-2019` will include golang.
i have used [this](https://github.com/actions/setup-go) for Linux VMs in workflows, but if it doesn't work the alternative is  manually download golang for windows from their website.

go.mod
```diff
@@ -0,0 +1,8 @@
+module sigs.kubernetes.io/sig-windows-tools
+
+go 1.14
+
+require (
+	github.com/Masterminds/semver v1.5.0
```
neolit123: this should work for us for now, but i haven't tried it for the k8s specific metadata (if we decide to build kube-proxy from master one day).

i know the apimachinery/pkg/util/version works for sure.

.github/workflows/kube-proxy.yml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  build:
+    runs-on: windows-2019
+    steps:
+    - uses: actions/checkout@v2
+    - uses: azure/docker-login@v1
+      with:
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_SECRET }}
```
neolit123: did you manage set these in the repository?

kubeadm/hack/publish-kubeproxy.go
```diff
@@ -0,0 +1,65 @@
...
+
+func dockerPush() error {
+	cmd := exec.Command("docker", "push", "sigwindowstools/kube-proxy")
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+func main() {
+	dockerfilePath := os.Args[1]
```
neolit123: this will panic if the user did not feed the path as an argument.

better to have a:
```
if len(os.Args) < 2 {
	// print usage
}
```
even for completely scoped tools.

kubeadm/hack/publish-kubeproxy.go
```diff
@@ -0,0 +1,65 @@
...
+		log.Fatalf("err fetching tags %v", err)
+	}
+	for _, t := range tags {
+		v, err := semver.NewVersion(t)
+		if err != nil {
+			log.Printf("err parsing %s, %v", t, err)
+		}
+		if constraint.Check(v) {
+			if err := dockerBuild(v, dockerfilePath); err != nil {
+				log.Fatalf("Error building docker image for %v: %v", v, err)
```
neolit123: errors should start lowercase in log messages.

kubeadm/hack/publish-kubeproxy.go
```diff
@@ -0,0 +1,65 @@
...
+			}
+		}
+	}
+	if err := dockerPush(); err != nil {
+		log.Fatalf("Error pushing docker image: %v", err)
```
neolit123: ditto.

.github/workflows/kube-proxy.yml
benmoss: It does: https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md#go-x64

kubeadm/hack/publish-kubeproxy.go
```diff
@@ -0,0 +1,65 @@
...
+func dockerPush() error {
+	cmd := exec.Command("docker", "push", "sigwindowstools/kube-proxy")
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+func main() {
+	dockerfilePath := os.Args[1]
+	constraint, err := semver.NewConstraint(supportedVersions)
```
neolit123: i have not tested this before.
given we don't have unit tests, hoping that we will not trigger a push for "all" kube-proxy images. :)

.github/workflows/kube-proxy.yml
neolit123: great!

go.mod
benmoss: Yeah, if we need something we can change later

.github/workflows/kube-proxy.yml
benmoss: Ah crud I don't have permission on the kubernetes-sigs repo, looks like I'll need to change myself to an admin to do so

kubeadm/hack/publish-kubeproxy.go
benmoss: Yeah, was just being lazy but I'll fix this

kubeadm/hack/publish-kubeproxy.go
benmoss: Worst case is still just a whole lot of images get pushed to docker hub, not a doomsday scenario. But it works in my tests, like I mentioned you can see the last run before I rebased and removed the push trigger here: https://github.com/benmoss/sig-windows-tools/actions/runs/53251162

neolit123:
/lgtm
@benmoss 
feel free to cancel the hold if you think this is in a good shape.

PR #54, opened 2020-03-16
Lower parallelism

benmoss:
hack/ginkgo-e2e [defaults to 25](https://github.com/kubernetes/kubernetes/blob/84dc7046797aad80f258b6740a98e79199c8bb4d/hack/ginkgo-e2e.sh#L128-L132), which we speculate is causing test failures because it's overloading our nodes

/assign @neolit123


benmoss:
Didn't notice #53 😸

PR #60, opened 2020-03-24
Update flannel to release 0.12.0

benmoss:
/assign @neolit123 

https://github.com/coreos/flannel/releases/tag/v0.12.0 came out recently so we can finally move off of my release binary. I already pushed the image to the Docker Hub account under a new tag, so I'll make a new release of sig-windows-tools after this is merged.


neolit123:
nice
/lgtm

PR #66, opened 2020-04-09
Create docker host nat network on demand

### LINKED ISSUE #52: Windows nodes do not come back after reboot
We disabled automatic reboots for our e2e clusters in #51, but we should make it so even if a machine reboots that kubelet / docker / wins all are set to automatically start so that a machine can be rebooted.

lippertmarkus:
Workaround, fixes #52 

Not really beautiful doing that inside the kubelet start script but it's the easiest solution. 

Another alternative would be to create another external script file and somehow set up something like the Windows task scheduler to run it on startup. But then we would also need to make sure that this script waits for startup of the docker service which adds additional complexity.


artisticcheese:
@lippertmarkus I'm confused by the script. Is not this script is used only once during node onboarding and hence will not be running on reboot?

PR #73, opened 2020-04-29
Fix source Vip endpoint creation failure

xmapst:
Fix source Vip endpoint creation failure, Allow Pods on Windows / Linux nodes to communicate with each other


xmapst:
@mortent

PR #74, opened 2020-04-30
remove the Alpha grade kubeadm integration (scripts / json)

neolit123:
The versioned k/website docs probably still reference this, but it's
unsupported by the maintainers.

/kind cleanup deprecation
/priority important-longterm
/assign @benmoss


ksubrmnn:
Is this so that we can move to the daemonset approach? That's fine, but perhaps we can keep both versions of scripts so that users can choose what suits them. There are customers who prefer services and customers who prefer daemonsets so let's do both if possible :)

neolit123:
> Is this so that we can move to the daemonset approach? 

yes

> That's fine, but perhaps we can keep both versions of scripts so that users can choose what suits them. There are customers who prefer services and customers who prefer daemonsets so let's do both if possible :)

i believe @benmoss mentioned to me that this could be the case. a couple of issues with that are:
- the old guide / docs were removed from k/website master with the DS approach. also the old docs will phase-out too with the k8s support skew (e.g. 1.16 is out of support with the release of 1.19).
- if sig-windows are willing to support and have the bandwidth to support both approaches that is fine, but they need to reside in separate folders and the service approach needs to be up-to-date with the k8s support skew.

benmoss:
I don't have strong opinions on this. I don't see why we have to offer any "support" on the old scripts, they'll be undocumented but if people want to keep using them I don't see why we can't leave them around.

neolit123:
ok, i will close the PR and leave the cleanup to the owners and for the future.
thanks.

Issue #81, opened 2020-05-09
Remove need to specify --interface to flannel setup


oe-hbk:
The --interface argument for flannel setup, which  calls the New-HnsNetwork powershell function does not seem to be necessary.

Should remove this from the flannel config files.   The main kubernetes, https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/adding-windows-nodes/   documentation  should also  change to remove the need to modify this argument.

PR #82, opened 2020-05-09
Remove the hardcoded --interface argument that gets fed into New-HnsNetwork

oe-hbk:
This argument seems to not be necessary, and creates difficulty in  environments where  you don't have a consistent device name for the default NIC across all your nodes.

Fixes: https://github.com/kubernetes-sigs/sig-windows-tools/issues/81

### DIFF
--- a/kubeadm/flannel/flannel-host-gw.yml
+++ b/kubeadm/flannel/flannel-host-gw.yml
@@ -32,7 +32,7 @@ data:
     cp -force /k/flannel/* /host/k/flannel/
     cp -force /kube-proxy/kubeconfig.conf /host/k/flannel/kubeconfig.yml
     cp -force /var/run/secrets/kubernetes.io/serviceaccount/* /host/k/flannel/var/run/secrets/kubernetes.io/serviceaccount/
-    wins cli process run --path /k/flannel/setup.exe --args "--mode=l2bridge --interface=Ethernet"
+    wins cli process run --path /k/flannel/setup.exe --args "--mode=l2bridge"
     wins cli route add --addresses 169.254.169.254
     wins cli process run --path /k/flannel/flanneld.exe --args "--kube-subnet-mgr --kubeconfig-file /k/flannel/kubeconfig.yml" --envs "POD_NAME=$env:POD_NAME POD_NAMESPACE=$env:POD_NAMESPACE"
   cni-conf.json: |
--- a/kubeadm/flannel/flannel-overlay.yml
+++ b/kubeadm/flannel/flannel-overlay.yml
@@ -31,7 +31,7 @@ data:
     cp -force /k/flannel/* /host/k/flannel/
     cp -force /kube-proxy/kubeconfig.conf /host/k/flannel/kubeconfig.yml
     cp -force /var/run/secrets/kubernetes.io/serviceaccount/* /host/k/flannel/var/run/secrets/kubernetes.io/serviceaccount/
-    wins cli process run --path /k/flannel/setup.exe --args "--mode=overlay --interface=Ethernet"
+    wins cli process run --path /k/flannel/setup.exe --args "--mode=overlay"
     wins cli route add --addresses 169.254.169.254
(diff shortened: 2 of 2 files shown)


neolit123:
please sign the CLA as the bot explained above. after that we can merge the PR if approved.

/assign @benmoss @ksubrmnn 
is this change appropriate?

benmoss:
This is going to thread through to the [setup.go](https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/kubeadm/flannel/setup.go#L36) script. I have no idea what interface will be selected if you let `New-HNSNetwork` "autodetect" your interface. This doesn't seem viable.

ksubrmnn:
I agree with @benmoss but I also understand the issues you're facing @oe-hbk 

I think the solution here is to make the interface argument optional and if it is not supplied we should automatically select the interface via powershell. 

@JocelynBerrendonner can help you with that :)

oe-hbk:
I was originally going to add something like:
Get-NetRoute -DestinationPrefix "0.0.0.0/0" | Get-NetAdapter -ErrorAction SilentlyContinue |Select-Object -ExpandProperty Name 

To get the interface name with the default route. But once I saw that it wasn't even required by the Hns calls,  I skipped it.   I will try to investigate how it determines the network in the absence of one provided, but for our environment, it's always picked the correct one so far.

JocelynBerrendonner:
+1 on Ben Moss's comment. IIRC, New-HnsNetwork will select the first physical interface connected to a network when it is not passed any interface parameter. If there are multiple physical interfaces connected to a network on the system, the selection will pretty much randomly pick one of these interfaces. The "interface" parameter is there to guarantee selection uniqueness.
Passing the name of the "default route" interface is a better option than not passing any interface name, but it becomes problematic when two such interfaces exist on the system. Metric could help, but there is no guarantee this results in the configuration users actually want. Passing the name is the only way to make the end result predictable.
@oe-hbk : Network interfaces can be renamed in Windows. As far as I recall, we suggest users to name their interfaces uniformly across their nodes. Is this an issue in your case?

oe-hbk:
@JocelynBerrendonner ,  yes, we have a mix of VMs and Physicals.   The Physicals have two NIC ports that are teamed together (new-netlbfoteam,  non-switch based)  It would be a bit odd to have the same name across these different types.  Doable, but not practical.
Would it make sense to have --interface be settable to "DEFAULT" or some other constant and hns.psm1 handle the logic to get the interface with default route (with lowest metric)

benmoss:
Hmm, maybe setup.go could also do that. Add a `--auto-interface` feature that if there's only one interface selects that one, or otherwise uses some logic to select one.

JocelynBerrendonner:
Agreed on adding an "--auto-interface" feature. That would be a good feature to add. I would keep the default behavior as-is, though.

oe-hbk:
Ok,  I'll update the PR to reflect that.  Thanks for your suggestions.

oe-hbk:
So I  gave this a shot, but didn't end up liking the results.  When I tried to get the network with the default route, 
`Get-NetRoute -DestinationPrefix "0.0.0.0/0" | Sort-Object -Property @{Expression = {$_.RouteMetric + $_.InterfaceMetric}; Descending = $False } | Select-Object -First 1 | Get-NetAdapter -ErrorAction SilentlyContinue | Select-Object -ExpandProperty Name`

It works great on different machines I've tried it out on.  But after the first run of the kubelet when it starts the flannel, the default NIC is no longer something like "Ethernet 0"   but now becomes a Hyper-V Virtual Ethernet Adapter with the name like  vEthernet (Ethernet 0).  And it's not any of the sorting going on above, the vethernet device becomes the "only" nic that has the default route.

So subsequent restarts of flannel, will now pick this device up as the auto-interface with this logic.  Which doesn't seem correct to me.

Does anyone know more about how this is supposed to work, and can tell me if the default route set by flannel to use the vEthernet device is expected?

Thanks

benmoss:
If there is already a `vEthernet (Ethernet*)` device can't we assume we already have created the HNS network and do nothing?

Issue #86, opened 2020-05-12
"--node-ip" param for kubelet is needed to be supported in PrepareNode.ps1


ruicao93:
For windows node with multiple interfaces or multuple IP addresses. It's needed to configure NodeIP used by kubelet in such case.  Currently `PrepareNode.ps1` script does not support configure the Node IP on kubelet. So I think we can add a param in  `PrepareNode.ps1` to support the config.

Kubelet receives ["--node-ip"](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet) for the setting.
```
Kubelet param:
--node-ip string
   IP address of the node. If set, kubelet will use this IP address for the node
```

ruicao93:
It seems that `kubelet` will retrive params from file `kubeadm-flags.env` at startup. So I think user can specify the `--node-ip` in this file for the NodeIP config.

By this way we don't need to change the `PrepareNode.ps1` or `StartKubelet.ps1` for this config.

I will close the PR and Issue.

```
$FileContent = Get-Content -Path "/var/lib/kubelet/kubeadm-flags.env"
$global:KubeletArgs = $FileContent.Trim("KUBELET_KUBEADM_ARGS=`"")
...
$cmd = "C:\k\kubelet.exe $global:KubeletArgs...
```
Here's the content of `kubeadm-flags.env`
```
cat /var/lib/kubelet/kubeadm-flags.env
KUBELET_KUBEADM_ARGS="--cgroup-driver= --network-plugin=cni --pod-infra-container-image=k8s.gcr.io/pause:3.1"
```

PR #87, opened 2020-05-12
Support specify Node IP used by kubelet

ruicao93:
Append "--node-ip" param in kubelet startup params. User can set
"NODE_IP" environment variable to configure the node IP used by kubelet.

Fixes: #86

### DIFF
--- a/kubeadm/scripts/PrepareNode.ps1
+++ b/kubeadm/scripts/PrepareNode.ps1
@@ -7,6 +7,7 @@ This script assists with joining a Windows node to a cluster.
 - Downloads Kubernetes binaries (kubelet, kubeadm) at the version specified
 - Registers wins as a service in order to run kube-proxy and cni as DaemonSets.
 - Registers kubelet as an nssm service. More info on nssm: https://nssm.cc/
+- User can set "NODE_IP" environment variable to configure the node IP used by kubelet.
 
 .PARAMETER KubernetesVersion
 Kubernetes version to download and use
@@ -72,8 +73,11 @@ $netId = docker network ls -f name=host --format "{{ .ID }}"
 if ($netId.Length -lt 1) {
     docker network create -d nat host
 }
-
-$cmd = "C:\k\kubelet.exe $global:KubeletArgs --cert-dir=$env:SYSTEMDRIVE\var\lib\kubelet\pki --config=/var/lib/kubelet/config.yaml --bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf --kubeconfig=/etc/kubernetes/kubelet.conf --hostname-override=$(hostname) --pod-infra-container-image=`"mcr.microsoft.com/oss/kubernetes/pause:1.3.0`" --enable-debugging-handlers --cgroups-per-qos=false --enforce-node-allocatable=`"`" --network-plugin=cni --resolv-conf=`"`" --log-dir=/var/log/kubelet --logtostderr=false --image-pull-progress-deadline=20m"
+$NodeIP = ""
+if ($env:NODE_IP -ne $null) {
+    $NodeIP = $env:NODE_IP
+}
+$cmd = "C:\k\kubelet.exe $global:KubeletArgs --cert-dir=$env:SYSTEMDRIVE\var\lib\kubelet\pki --config=/var/lib/kubelet/config.yaml --bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf --kubeconfig=/etc/kubernetes/kubelet.conf --hostname-override=$(hostname) --pod-infra-container-image=`"mcr.microsoft.com/oss/kubernetes/pause:1.3.0`" --enable-debugging-handlers --cgroups-per-qos=false --enforce-node-allocatable=`"`" --network-plugin=cni --resolv-conf=`"`" --log-dir=/var/log/kubelet --logtostderr=false --image-pull-progress-deadline=20m --node-ip=$NodeIP"
 
 Invoke-Expression $cmd'
 Set-Content -Path $global:StartKubeletScript -Value $StartKubeletFileContent


kubeadm/scripts/PrepareNode.ps1
```diff
@@ -46,6 +47,7 @@ $kubeletBinPath = "$global:KubernetesPath\kubelet.exe"
 mkdir -force "$global:KubernetesPath"
 $env:Path += ";$global:KubernetesPath"
 [Environment]::SetEnvironmentVariable("Path", $env:Path, [System.EnvironmentVariableTarget]::Machine)
+[Environment]::SetEnvironmentVariable("NODE_IP", $NodeIP, [System.EnvironmentVariableTarget]::Machine)
```
neolit123, line 50: instead of taking an flag to the ps1 script, which then sets an env. variable which is then passed to the kubelet, why not pass the `--node-ip=....` to the kubelet only if the user has set `NODE_IP` outside of the ps1 - e.g. in their environment?

neolit123: or just pass the value of the ps1 flag `NodeIP` directly to `--node-ip` to the kubelet.
if `--node-ip` equals `""` the kubelet should apply it's detection.

ruicao93: Hi @neolit123 ,
- Firstly, consider the kubelet runs as service and user is easy to update `NODE_IP` by setting environment. So I think get the param from environment when booting kubelet is good.
- Basic on the above choice, ps1 just need to pass the flag NodeIP to environment. If we pass the flag to kubelet directly, user will must change the  `StartKubelet.ps1` when they want to change the node ip.

How do you think?

neolit123: ok, my vote is to go with the first option in this PR.

but also please add a note under `.DESCRIPTION`:
> You can use set the "NODE_IP" environment variable to configure the node IP for the kubelet

ruicao93: Thanks Lubomir. Update the `.DESCRIPTION` per your comment.

neolit123: but i think this is not needed anymore, right?
```
[Environment]::SetEnvironmentVariable("NODE_IP", $NodeIP, [System.EnvironmentVariableTarget]::Machine)
```
because the user is supposed to set the env variable now.

neolit123:
please keep the commits squashed to 1.

kubeadm/scripts/PrepareNode.ps1
ruicao93: Yes, correct. Remove the flag from ps1.

ruicao93:
> please keep the commits squashed to 1.

Done.

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -73,7 +74,7 @@ if ($netId.Length -lt 1) {
     docker network create -d nat host
 }
 
-$cmd = "C:\k\kubelet.exe $global:KubeletArgs --cert-dir=$env:SYSTEMDRIVE\var\lib\kubelet\pki --config=/var/lib/kubelet/config.yaml --bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf --kubeconfig=/etc/kubernetes/kubelet.conf --hostname-override=$(hostname) --pod-infra-container-image=`"mcr.microsoft.com/oss/kubernetes/pause:1.3.0`" --enable-debugging-handlers --cgroups-per-qos=false --enforce-node-allocatable=`"`" --network-plugin=cni --resolv-conf=`"`" --log-dir=/var/log/kubelet --logtostderr=false --image-pull-progress-deadline=20m"
+$cmd = "C:\k\kubelet.exe $global:KubeletArgs --cert-dir=$env:SYSTEMDRIVE\var\lib\kubelet\pki --config=/var/lib/kubelet/config.yaml --bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf --kubeconfig=/etc/kubernetes/kubelet.conf --hostname-override=$(hostname) --pod-infra-container-image=`"mcr.microsoft.com/oss/kubernetes/pause:1.3.0`" --enable-debugging-handlers --cgroups-per-qos=false --enforce-node-allocatable=`"`" --network-plugin=cni --resolv-conf=`"`" --log-dir=/var/log/kubelet --logtostderr=false --image-pull-progress-deadline=20m --node-ip=$env:NODE_IP"
```
neolit123, line 77: hi, just to double check. PowerShell will return "" here in case the user has not set the `NODE_IP` env. var, correct?

benmoss: Yeah, it will. What does `--node-ip=""` do?

neolit123: tells the kubelet to pick an IP from one of the interfaces on the node. on Linux at least, this usually ends up being the external one.

it's only documented with source code comments.

ruicao93: From my test Powershell will return NULL if `NODE_IP` is not set. But I think kubelet can still startup with param `--node-ip=`.

neolit123: does it return "NULL" as a string?
if you pass "NULL" to the kubelet it would fail; it must be "".

ruicao93: > does it return "NULL" as a string?
> if you pass "NULL" to the kubelet it would fail; it must be "".

Nope, it just retun a NULL value, neither "NULL" in string or empty string. The param will be `--node-ip=`.

neolit123: so if it's `--node-ip=` it will break...

i propose to add a PowerShell `if` check above the kubelet call to optionally add the whole `--node-ip=value` only if the envvar NODE_IP is set.

neolit123: pseudo code:
```
var nodeIPFlag = ""
if env(NODE_IP) != "" {
   nodeIPFlag="--node_ip="+env(NODE_IP)
}
kubelet .... $nodeIPFlag
```

ruicao93: @neolit123 , It seems that `kubelet` will retrive params from file `kubeadm-flags.env`. So I think user can specify the `--node-ip` in this file for the NodeIP config.

By this way we don't need to change the `PrepareNode.ps1` or `StartKubelet.ps1` for this config.

I will close the PR and Issue. Thanks @neolit123  and @benmoss a lot for your review.

```
$FileContent = Get-Content -Path "/var/lib/kubelet/kubeadm-flags.env"
$global:KubeletArgs = $FileContent.Trim("KUBELET_KUBEADM_ARGS=`"")
...
$cmd = "C:\k\kubelet.exe $global:KubeletArgs...
```
Here's the content of `kubeadm-flags.env`
```
cat /var/lib/kubelet/kubeadm-flags.env
KUBELET_KUBEADM_ARGS="--cgroup-driver= --network-plugin=cni --pod-infra-container-image=k8s.gcr.io/pause:3.1"
```

ruicao93:
It seems that `kubelet` will retrive params from file `kubeadm-flags.env` at startup. So I think user can specify the `--node-ip` in this file for the NodeIP config.

By this way we don't need to change the `PrepareNode.ps1` or `StartKubelet.ps1` for this config.

I will close the PR and Issue.

```
$FileContent = Get-Content -Path "/var/lib/kubelet/kubeadm-flags.env"
$global:KubeletArgs = $FileContent.Trim("KUBELET_KUBEADM_ARGS=`"")
...
$cmd = "C:\k\kubelet.exe $global:KubeletArgs...
```
Here's the content of `kubeadm-flags.env`
```
cat /var/lib/kubelet/kubeadm-flags.env
KUBELET_KUBEADM_ARGS="--cgroup-driver= --network-plugin=cni --pod-infra-container-image=k8s.gcr.io/pause:3.1"
```

neolit123:
long term in kubeadm we are trying to deprecate / remove the `kubeadm-flags.env` file.
this file is dynamically generated by kubeadm join / init and should not be edited by the user.

ruicao93:
> long term in kubeadm we are trying to deprecate / remove the `kubeadm-flags.env` file.
> this file is dynamically generated by kubeadm join / init and should not be edited by the user.

Thanks @neolit123 for your explanation. 

We're preparing [Antrea CNI](https://github.com/vmware-tanzu/antrea/issues/331) support for windows. So do you think if we should use `kubeadm-flags.env` file in current stage or continue to push this PR?

Thanks,
Rui

benmoss:
I wouldn't use kubeadm-flags.env, it's not meant to be edited. I think PrepareNode.ps1 is the right place to do this.

benmoss:
Yeah, I'm not sure. It seems people do the same thing on Linux, so maybe it's okay for now. https://github.com/kubernetes/kubeadm/issues/203

benmoss:
We don't have as many features built around having separate files for kubeadm-set variables vs user-set variables as are on Linux, but we're also not writing a service definition file like they do for Linux. There the correct way to do this is with `/etc/default/kubelet` or `/etc/sysconfig/kubelet` which populates the `KUBELET_EXTRA_ARGS` variable. We could introduce something like that but I think it might be overkill.

PR #106, opened 2020-10-16
Check that HNS network doesn't exist before creating new one

vitaliy-leschenko:
When we restart flannel pod we can see error message like this:
```
Invoke-HnsRequest : @{Error=A network with this name already exists. ; ErrorCode=2151350288; Success=False}
At C:\k\flannel\hns.psm1:233 char:16
+ ...      return Invoke-HnsRequest -Method POST -Type networks -Data $Json ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,Invoke-HNSRequest
```
because HNS network 'External' already exists.

### DIFF
--- a/kubeadm/flannel/setup.go
+++ b/kubeadm/flannel/setup.go
@@ -10,7 +10,7 @@ import (
 
 func setupOverlay(interfaceName string) {
 	run(fmt.Sprintf(`ipmo C:\k\flannel\hns.psm1; `+
-                `$network = Get-HNSNetwork | ? Name -eq "External" `+
+                `$network = Get-HNSNetwork | ? Name -eq "External"; `+
                 `if ($network -eq $null) { `+
                 `New-HNSNetwork -Type Overlay -AddressPrefix "192.168.255.0/30" -Gateway "192.168.255.1" -Name "External" -AdapterName "%s" -SubnetPolicies @(@{Type = "VSID"; VSID = 9999; }); `+
                 `} elseif ($network.Type -ne "Overlay") { `+
--- a/kubeadm/flannel/setup.go
+++ b/kubeadm/flannel/setup.go
@@ -10,18 +10,24 @@ import (
 
 func setupOverlay(interfaceName string) {
 	run(fmt.Sprintf(`ipmo C:\k\flannel\hns.psm1; `+
-                `if ((Get-HNSNetwork | ?{ ($_.Type -eq "Overlay") -and ($_.Name -eq "External") }) -eq $null) { `+
-                `New-HNSNetwork -Type Overlay -AddressPrefix "192.168.255.0/30" `+
-		`-Gateway "192.168.255.1" -Name "External" -AdapterName "%s" -SubnetPolicies @(@{Type = "VSID"; VSID = 9999; }); }`,
+                `$network = Get-HNSNetwork | ? Name -eq "External" `+
+                `if ($network -eq $null) { `+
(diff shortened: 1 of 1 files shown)


vitaliy-leschenko:
I signed it

kubeadm/flannel/setup.go
```diff
@@ -9,15 +9,19 @@ import (
 )
 
 func setupOverlay(interfaceName string) {
-	run(fmt.Sprintf(`ipmo C:\k\flannel\hns.psm1; New-HNSNetwork -Type Overlay -AddressPrefix "192.168.255.0/30"`+
-		`-Gateway "192.168.255.1" -Name "External" -AdapterName "%s" -SubnetPolicies @(@{Type = "VSID"; VSID = 9999; })`,
+	run(fmt.Sprintf(`ipmo C:\k\flannel\hns.psm1; `+
+                `if ((Get-HNSNetwork | ?{ ($_.Type -eq "Overlay") -and ($_.Name -eq "External") }) -eq $null) { `+
```
jsturtevant, line 13: If there is a network that is not overlay  with same name you will end up with the same error since it will try to create the network aftwards.  Someone might also have a bad day if that network named External is not of type "Overlay".  

Could you warn if the network type is not overlay? It might make sense to pull this out to a constant or use go templates to fill the values?

jsturtevant (changes requested):
Thanks for the contribution!

kubeadm/flannel/setup.go
vitaliy-leschenko: Thank for review.
I added check for exist network type and warning about wrong type.
Please review again.

vitaliy-leschenko: > It might make sense to pull this out to a constant or use go templates to fill the values?
Maybe you are right but I have no go skills and tried to make minimal changes.

kubeadm/flannel/setup.go
```diff
@@ -9,15 +9,25 @@ import (
 )
 
 func setupOverlay(interfaceName string) {
-	run(fmt.Sprintf(`ipmo C:\k\flannel\hns.psm1; New-HNSNetwork -Type Overlay -AddressPrefix "192.168.255.0/30"`+
-		`-Gateway "192.168.255.1" -Name "External" -AdapterName "%s" -SubnetPolicies @(@{Type = "VSID"; VSID = 9999; })`,
+	run(fmt.Sprintf(`ipmo C:\k\flannel\hns.psm1; `+
+                `$network = Get-HNSNetwork | ? Name -eq "External" `+
```
jsturtevant, line 13:
```suggestion
                `$network = Get-HNSNetwork | ? Name -eq "External"; `+
```

kubeadm/flannel/setup.go
```diff
@@ -9,15 +9,25 @@ import (
...
+                `}`,
 		interfaceName),
 	)
 }
 
 func setupL2bridge(interfaceName string) {
-	run(fmt.Sprintf(`ipmo C:\k\flannel\hns.psm1; New-HNSNetwork -Type Overlay -AddressPrefix "192.168.255.0/30"`+
-		`-Gateway "192.168.255.1" -Name "External" -AdapterName "%s"`,
+	run(fmt.Sprintf(`ipmo C:\k\flannel\hns.psm1; `+
+                `$network = Get-HNSNetwork | ? Name -eq "External" `+
```
jsturtevant, line 25:
```suggestion
                `$network = Get-HNSNetwork | ? Name -eq "External"; `+
```

jsturtevant:
Thanks. Just missing a closing semi colon

jsturtevant (approved):

PR #109, opened 2020-10-23
Update flannel docker image to 0.13.0

vitaliy-leschenko:
Flannel 0.13.0 has been released. Maybe make sense build new image `sigwindowstools/flannel:0.13.0` to be consistent with it.

When new version of docker image ready we need to update yaml files:
- flannel-host-gw.yml
- flannel-overlay.yml

### DIFF
--- a/kubeadm/flannel/Dockerfile
+++ b/kubeadm/flannel/Dockerfile
@@ -13,7 +13,7 @@ ARG cniVersion
 
 RUN mkdir -force C:\k\flannel; \
   pushd C:\k\flannel; \
-  curl.exe -LO https://github.com/coreos/flannel/releases/download/v0.12.0/flanneld.exe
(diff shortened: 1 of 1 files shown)


jsturtevant:
looks fine to me, Have you tried this out in a cluster? We don't have automation for building new images right now.  Created #110 to track

vitaliy-leschenko:
I tried to use it in my cluster. It work fine.

Also I tried to change base image from servercore to nanoserver. It also work well enough. I want to create another PR for it. Maybe make sense to combine it. What do you think about it?

jsturtevant:
let's close this in preference of #111 which uses args: https://github.com/kubernetes-sigs/sig-windows-tools/pull/111/files#diff-636021326186ef241842669769c8f799af4c114c5192522d0252d88a937c5362

Issue #110, opened 2020-10-23
Add automation for building kubeadm related containers


jsturtevant:
There is currently a github action for building kube-proxy: https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/.github/workflows/kube-proxy.yml

There should be a pipeline for building and publishing flannel images as well.  Should be able to use kube-proxy as an example

vitaliy-leschenko:
I will try to do it

benmoss:
Yeah, I think the only reason we didn't do this was because flannel releases once in a blue moon (aka very rarely)

vitaliy-leschenko:
@jsturtevant please review my PR to fix the issue

jsturtevant:
fixed by #111

PR #111, opened 2020-10-23
Automation for building flannel and kubeproxy related containers

vitaliy-leschenko:
Fix issues #110, #123 

- [x] Change base image to nanoserver
- [x] Update docker files (flannel + kubeproxy) to use buildx to support multiple Windows OS
- [x] Add action on PR for build (only build):
 - `sigwindowstools/flannel:{flannelVersion}-{windowsVersion}` - images for windows server 1809, 1903, 1909, 2004
 - `sigwindowstools/kube-proxy:{k8sVersion}-{windowsVersion}` - images for windows server 1809, 1903, 1909, 2004 for each k8s version >= 1.17.0
- [x] Add action on push to master for build and push:
 - `sigwindowstools/flannel:{flannelVersion}-{windowsVersion}` - images for windows server 1809, 1903, 1909, 2004
 - `sigwindowstools/flannel:{flannelVersion}` - multi arch manifest based on all previous images
 - `sigwindowstools/kube-proxy:{k8sVersion}-{windowsVersion}` - images for windows server 1809, 1903, 1909, 2004 for each k8s version >= 1.17.0
 - `sigwindowstools/kube-proxy:{k8sVersion}` - multi arch manifests based on all previous images
- [x] Update `flannel-host-gw.yml` and `flannel-overlay.yml` to use `sigwindowstools/flannel:0.13.0`
- [x] Update `kube-proxy.yml`


kubeadm/hack/publish-flannel.go
```diff
@@ -0,0 +1,69 @@
...
+		fmt.Sprintf("--tag=sigwindowstools/flannel:%s", v.String()),
+		dockerfilePath,
+	)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+func dockerPush() error {
+	cmd := exec.Command("docker", "push", "sigwindowstools/flannel")
```
neolit123, line 30: Just noting that with some of the inbound dockerhub limitations we may have to migrate these to someplace else.

.github/workflows/flannel.yml
```diff
@@ -0,0 +1,17 @@
+name: kube-proxy-images
```
neolit123, line 1: Should be flannel-images

neolit123 (commented):
Thanks

.github/workflows/flannel.yml
vitaliy-leschenko: You are right. I will fix it

kubeadm/hack/publish-flannel.go
marosset: Noted.
@ddebroy was filing some exceptions to the inbound dockerhub limitations for the registries most of the e2e test images are stored (mainly e2eteam) since these are open source projects.
We should see if we can get an exception for sigwindowstools registry as well.

vitaliy-leschenko: This code similar to publish-kubeproxy.go

.github/workflows/flannel.yml
```diff
@@ -0,0 +1,17 @@
+name: flannel-images
+
+on:
+  schedule:
+  - cron: "0 0 * * *"
```
jsturtevant, line 5: I don't think we will want this one on a schedule since flannel doesn't need updates every version. We can publish a new version with a PR merges (push to master in github workflow terms) and when we cut a release?

https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#push 
https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#release

.github/workflows/flannel.yml
```diff
@@ -0,0 +1,17 @@
...
+  build:
+    runs-on: windows-2019
+    steps:
+    - uses: actions/checkout@v2
+    - uses: azure/docker-login@v1
+      with:
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_SECRET }}
+    - name: Build
+      run: go run kubeadm/hack/publish-flannel.go kubeadm/flannel
```
jsturtevant, line 17: Is there a reason we a binary to build and publish the image over just running docker build/push commands?

.github/workflows/flannel.yml
jsturtevant: We can tag the image version as pre release for merge to master and use the git tag for the release.

jsturtevant (changes requested):
Thanks taking a stab at this!  I am not sure the binary is needed for this use case. I believe it was added to automatically look up Kubernetes releases and publish new kube-proxy. In this case flannel doesn't change often and it's more likely a fix in our files that would cause a new image (like in #106).

.github/workflows/flannel.yml
vitaliy-leschenko: As I understand it required to build docker images with tags similar to flannel versions.
Anyway this file is modified version of `.github/workflows/kube-proxy.yml`

.github/workflows/flannel.yml
vitaliy-leschenko: So, how I can to resolve this moment? Do I need to remove `.github/workflows/flannel.yml`?

.github/workflows/flannel.yml
jsturtevant: I believe it was added to automatically look up Kubernetes releases and publish new kube-proxy. In this case flannel doesn't change often and it's more likely a fix in our files that would cause a new image (like in #106) so I think we don't need this extra look up

.github/workflows/flannel.yml
jsturtevant: use the docs above to add pr and release triggers.  an example of building on PR merge:

```
on:
  push:
    branches:
      - master
```

vitaliy-leschenko: ok

.github/workflows/flannel.yml
vitaliy-leschenko: Do I need to do any additional changes except trigger settings?

vitaliy-leschenko: If we change the code `go run kubeadm/hack/publish-flannel.go kubeadm/flannel` to `docker build/docker push` then we need to update the code after each new flannel version.
I changed workflow trigger as you said. So flannel images will be updated only after future commit in master.

kubeadm/hack/publish-flannel.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+const (
+	supportedVersions = ">= 0.12.0"
+)
+
+func dockerBuild(v *semver.Version, dockerfilePath string) error {
+	cmd := exec.Command("docker", "build", "--pull",
+		fmt.Sprintf("--build-arg=flannelVersion=%s", v.String()),
+		fmt.Sprintf("--tag=sigwindowstools/flannel:%s", v.String()),
```
jsturtevant, line 21: We should add our own version number appended to flannel.  Otherwise when we change files not related to the flannel version we will overwrite the tag with the newer version.

.github/workflows/flannel.yml
jsturtevant: This will build and push for each version greater than 0.12? from that perspective I think it would be useful to build each of the versions with new scripts. So I guess I am convinced. :-)

.github/workflows/flannel.yml
```diff
@@ -0,0 +1,18 @@
+name: flannel-images
+
+on:
+  push:
+    branches:
+      - master
```
jsturtevant, line 6: Can we also do a docker file build on PR submission? That would help ensure the dockerfilie works

jsturtevant:
@vitaliy-leschenko Thanks for the effort here! I left a few more comments.

kubeadm/hack/publish-flannel.go
vitaliy-leschenko: Maybe you are right. I don't know what is better: build huge numbers of tags (one for future merge) or replace tags by updated version. For example, base windows tags replace previous tags when rollout new ones.

So, maybe it is not so bad to replace previous tags.

.github/workflows/flannel.yml
vitaliy-leschenko: We can replace `go` application to many Dockerfiles. Just for example:
- Dockerfile.0.12.0-servercore1809
- Dockerfile.0.13.0-servercore1809
- Dockerfile.0.12.0-nanoserver1809
- Dockerfile.0.13.0-nanoserver1809

When flannel rollout new version we just add new Dockerfiles

vitaliy-leschenko: Now we can build Dockerfile (/kubeadm/flannel/Dockerfile) with special argument to test it:
```
docker build -t sigwindowstools/flannel:0.13.0 --build-arg flannelVersion="0.13.0" .
```

.github/workflows/flannel.yml
vitaliy-leschenko: Yep. This will build tags for each flannel version greater than 0.12 and replace previous with same version.
This is the same situation as for kube-proxy. If we change Dockerfile for it next day all tags will be replaced.

vitaliy-leschenko:
@jsturtevant Thanks for review. I did't do any changed. I just tried to answer to you comments. Please review.

.github/workflows/flannel.yml
jsturtevant: We don't need separate dockerfiles if you use ARG BASEIMAGE and a simple script or build template.  See The pause image for an example. We don't need quite as complex scenario but shows how to use a single build file for multiple OS versions

https://github.com/kubernetes/kubernetes/blob/9a57dc20a98dcc7a6b2bd1323d4e27f84477cd1b/build/pause/Dockerfile_windows

kubeadm/hack/publish-flannel.go
jsturtevant: The base windows tags don't replace the previous ones. They act as "latest" which point to specific image layer. Having a tag for each would make it so we are tied to a release and specific set of files.  We can also tag a latest to make it easier for on-boarding.

.github/workflows/flannel.yml
vitaliy-leschenko: Ok. So we can use Dockerfile from this PR to build different tags for different flannel versions because it based on ARG flannelVersion.

Or I didn't correctly understand your first comment from this thread.

marosset: I think James is asking to update the triggers to look like

```yaml
on:
  push:
    branches:
      - main
  pull_request:
    branches:
      - main
```

So we can make sure PRs don't accidently break the docker builds.

marosset: We probably shouldn't publish the image for PR builds tho

vitaliy-leschenko: If we add trigger on PR it may break (replace) stable docker tags.

vitaliy-leschenko: @jsturtevant I mean that `mcr.microsoft.com/windows/servercore:ltsc2019` replaces previous `mcr.microsoft.com/windows/servercore:ltsc2019` when some patch rolled out for it. And this images have no `latest` tag.

Dockerfile in this PR have arguments to build images for different version of flannel. `publish-flannel.go` use it to build tags for all flannel version >= 0.12.
- sigwindowstools/flannel:0.12.0
- sigwindowstools/flannel:0.13.0

This tags will replace previous when we change Dockerfile.

jsturtevant: > We probably shouldn't publish the image for PR builds tho

Agreed, we should only build (not push) the docker file to give a signal that it works.  

> replaces previous mcr.microsoft.com/windows/servercore:ltsc2019 when some patch rolled out for it. And this images have no latest tag.

`ltsc2019` is acting as the tag `latest` but there is still a tagged version and they both point to the same layer. Run docker `run --rm jess/reg:v0.15.8 tags mcr.microsoft.com/windows/servercore | sort -ur` and then inspect manifests like `docker manifest inspect --verbose mcr.microsoft.com/windows/servercore:ltsc2019` and `docker manifest inspect --verbose mcr.microsoft.com/windows/servercore:10.0.17763.1518` to see this.

We should tag a specific version for folks who want to use a specific version (this should really be done in production) and tag a `flannel:0.12.0` for usability.

vitaliy-leschenko: Ok. I will remove `dockerPush` from publish-flannel.go

kubeadm/hack/publish-flannel.go
vitaliy-leschenko: Ok. I am going to add special suffix to image sigwindowstools/flannel.
For example: sigwindowstools/flannel:%flannelversion%-build%number%

vitaliy-leschenko:
@jsturtevant, I updated code based on your comments. Do you have another ones?

kubeadm/flannel/Dockerfile
```diff
@@ -1,4 +1,5 @@
 ARG servercoreTag="ltsc2019"
+ARG flannelVersion="0.13.0"
 ARG cniVersion="0.8.5"
```
fourpastmidnight, line 3: There is a newer version of CNI, 0.8.7. Should any of these be defaulted? Or should these instead have no default value, and if no value is provided, just reach out to Github to ascertain the latest release version and use that instead? I did this in my own script module that I've successfully used to setup a Windows Server 2019 2004 server as a worker node for some of the dependencies, such as `yq` and `wins`. (I just haven't gotten around to "backporting" that to the other dependencies, such as Flannel and CNI in the docker images.)

fourpastmidnight:
What I don't see here is updating or somehow "parameterizing" the `flannel-*.yml` and `kube-proxy.yml` files to use the newly built images. (Forgive me, I'm a bit new to some of this, so if this is off-base, be kind. 😉 I'm learning still.) As far as I understand the state of things, these YAML files are hard-coded to use images which are built on Windows Server 2019 1809 only. So anything other than that version, and the Windows Server node will fail to join to the cluster&mdash;probably with something along the lines of the node being stuck in ContainerCreating with log files indicating the image could not be started due to a version mismatch.

In my script module at https://github.com/fourpastmidnight/kubewin, on lines 1975 through 2009, I modify those files accordingly. (I did not have a local docker registry setup for storing my created images--they were built locally on the node to be joined to the cluster--so I also added the `imagePullPolicy: never` option. But for a production script, either this is omitted entirely, or it's left as an option to set this (for dev environments where a docker registry may not be available, as in my case)?

fourpastmidnight:
A few more comments, regarding `PrepareNode.ps1`. I found the code to configure and add Kubelet as a service did not work correctly.

I created two issues for it (unfortunately, GitHub mostly helpfully, but sometimes unhelpfully in this case, remembers previous entries put into the New Issue form--and my issues got garbled. So I'll relate them here since they affect this work.

First, is this:

https://github.com/kubernetes-sigs/sig-windows-tools/blob/8f6890d6fc6e36aff229c4a895a958576f297a77/kubeadm/scripts/PrepareNode.ps1#L68

This line is incorrect and leaves behind a trailing `"`.  Subsequently, when this line is executed:

https://github.com/kubernetes-sigs/sig-windows-tools/blob/8f6890d6fc6e36aff229c4a895a958576f297a77/kubeadm/scripts/PrepareNode.ps1#L76

there are additional parameters passed after those in `$global:KubeletArgs`. And because of the trailing `"`, those arguments will fail to be passed correctly, causing Kubelet to not start on Windows. Line 68 _should_ be:

```powershell
$global:KubeletArgs = $FileContent.Trim("KUBELET_KUBEADM_ARGS=") -replace '^"(.*)"$', '$1'
```

Secondly, I found the configuration parameters to Kubelet on line 76 above to be incorrect. Passing the `--config` parameter is not necessary, and in fact, prevents me from being able to run Kubelet successfully. It's not required, as during the join process (if I understand it correctly), the required configuration is downloaded from the master (or generated on the client???)--either way, the right thing happens and "it just works" without specifying `--config`.

Third, and lastly, if using a version of Windows Server that is not 1809 (LTSC 2019), then you also need to build a custom 'pause' infrastructure image. And this image needs to be passed to the Kubelet service as a configuration parameter (`--pod-infra-container-image`).

All of my images are built in my custom script module (lines 1857 through 1956&mdash;and yes, that section of code needs to be split out and refactored. I _only_ just got this all working _yesterday_! 😄 ) in `New-WindowsKubernetesClusterNode`, along with using a few helper cmdlets: `New-GoLangContainerImage`, `New-KubernetesFlannelDaemonSetContainerImage`, `New-KubernetesKubeProxyDaemonSetContainerImage` and `New-KubernetesPauseContainerImage`. (Sorry for the long cmdlet names, but I was trying to keep things organized as much by cmdlet name as by the cmdlets themselves.)

jsturtevant:
@fourpastmidnight 

>A few more comments, regarding PrepareNode.ps1. 

Looks like you created issues for those already. Let's keep comments and fixes related those on the issues so we don't get off track on the PR here. Thanks for your contributions with those. Some good finds.

> What I don't see here is updating or somehow "parameterizing" the flannel-*.yml and kube-proxy.yml files to use the newly built images.

If we use a manifest for the `flannel:0.12` and multi arch builds then the version would be automatically picked up.

jsturtevant:
@vitaliy-leschenko 

Couple comments, Sorry if I wasn't clear, this should probably been in the initial issue.

- I was suggesting we do a build for PR's and a Push for merge to master.  I think this can be accomplished using [expression](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions) to only push when merging to master. 
- I think we need to add a `ARG BASEIMAGE` to the Docker file and build it for each OS version (2019/1909/2004). This would add a little more logic to the go file or we could use a simple configuration file and a script in the workflow file its self. I perfer dropping the go file since it add some complexity in building.  
- These uses buildx to do this:
  - an example of the args that can be passed https://github.com/kubernetes/kubernetes/blob/9a57dc20a98dcc7a6b2bd1323d4e27f84477cd1b/build/pause/Makefile#L133   
  - the fields in the docker file here: https://github.com/kubernetes/kubernetes/blob/9a57dc20a98dcc7a6b2bd1323d4e27f84477cd1b/build/pause/Dockerfile_windows#L36-L37 
  - an example of a manifest generation: https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/1fa6ffe3335ac5f0f2c2ad34be37a06da2863a2e/docker/build.sh#L103

It would look like:

- Dockerfiles for versions 2019/1909/2004 and flannel versions 12+ are built but not pushed
- On merge to master the images are built and pushed. 

There would be the following images:

- sigwindowstools/flannel:v0.12 (multi arch manifest that points to the latest of the following images)
- sigwindowstools/flannel:v0.12-2019-{build_tag}
- sigwindowstools/flannel:v0.12-1909-{build_tag}
- sigwindowstools/flannel:v0.12-2004-{build_tag}

- sigwindowstools/flannel:v0.13 (multi arch manifest that points to the latest of the following images)
- sigwindowstools/flannel:v0.13-2019-{build_tag}
- sigwindowstools/flannel:v0.13-1909-{build_tag}
- sigwindowstools/flannel:v0.13-2004-{build_tag}

Hope that helps. If need more explanation or it's a bit to much let me know and I can take a try at it.

fourpastmidnight:
So, would the images be rebuilt every time there is a Cumulative Update (which changes the Windows Server version)? I think that has also been a source of frustration/bewilderment to those not expecting it? (Afterall, we _do_ want to be keeping the OS updated with security patches and whatnot, right?) I.e., I saw someone attempting to build a Kubernetes cluster with Windows Server 2019 2004 with version 10.0.19041.**388**, whereas, I'm using Windows Server 2019 2004 with version 10.0.19041.**508**. As far as I understand things, images would be incompatible between these two images??

fourpastmidnight:
One more comment, we should be using flannel 0.13.0. It's my understanding that there's a bug in flannel 0.12.0 that's been causing some headaches for people. I don't recall where I saw that info, nor the issue--but it was specifically with Windows, and I did see mention that it was resolved with flannel 0.13.0.

vitaliy-leschenko:
I added new powershell script to build flannel images for many flannel versions and windows versions.
You can see result of it: https://hub.docker.com/repository/docker/vleschenko/flannel/tags?page=1

How to use it:
```
cd kubeadm\flannel
.\build.ps1 -Image sigwindowstools/flannel -Push
```
But it requires build server WS2004+

vitaliy-leschenko:
Unfortunately github doesn't support windows server 2004 for build server.
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

kubeadm/flannel/versions.json
```diff
@@ -0,0 +1,22 @@
...
+    "flannel": [
+        "0.12.0", 
+        "0.13.0"
+    ],
+    "baseimages": [
+        {
+            "dockerfile": "Dockerfile",
+            "suffix": "",
+            "base": "mcr.microsoft.com/windows/servercore",
+            "tags": ["1809"],
```
vitaliy-leschenko, line 11: Currently github doesn't support Windows Server 1903 or 2004 as build server.
So if we add 1903 and 2004 to tags then build will be failed because it will be run on Windows Server 1809.
```
tags: ["1809", "1903", "2004"]
```

vitaliy-leschenko: We need to add self-hosted runner based on Windows Server 2004 to build images for Windows Server 2004.

jsturtevant: :disappointed:   might need to look more seriously at buildx. It is how the test images and upstream pause image are built.

vitaliy-leschenko: Sorry, maybe I am so stupid, but I don't understand how to use buildx on windows. It doesn't work for me.

jsturtevant: your not, it doesn't.  We would have to switch to building on Linux and requires no 'RUN' commands in the windows docker images. We should be able to pull this off but probably makes sense to do in a follow up PR

jsturtevant:
@vitaliy-leschenko looks like the build was triggered!

Screenshot from 2020-11-04 09-50-22

vitaliy-leschenko:
Yes, build was triggered but it doesn't build for 1903, 2004. It build only for 1809.

kubeadm/flannel/Dockerfile
jsturtevant: We could update this to .8.7

.github/workflows/flannel.yml
```diff
@@ -0,0 +1,30 @@
...
```
jsturtevant, line 6: Since this is only for the flannal images we should restrict these to that folder. https://github.blog/changelog/2019-09-30-github-actions-event-filtering-updates/

kubeadm/flannel/build.ps1
```diff
@@ -0,0 +1,43 @@
+param(
```
jsturtevant, line 1: Will this script work in powershell core?

kubeadm/flannel/Dockerfile.nanoserver
```diff
@@ -0,0 +1,35 @@
+ARG tag="nanoserver-1809"
```
jsturtevant, line 1: Moving this to nano is a good improvement. Is there any reason to keep the windowsservercore image?

jsturtevant:
>You can see result of it: https://hub.docker.com/repository/docker/vleschenko/flannel/tags?page=1

looking great!  I had just a few more comments and small changes.  Could you also re-base the changes?

kubeadm/flannel/Dockerfile
vitaliy-leschenko: Ok

kubeadm/flannel/build.ps1
vitaliy-leschenko: Yes, it will. I checked it on powershell core.

.github/workflows/flannel.yml
vitaliy-leschenko: Ok

kubeadm/flannel/Dockerfile.nanoserver
vitaliy-leschenko: I don't know. Maybe not. Nano is faster and smaller.
Do you think we need to remove servercore?

vitaliy-leschenko:
@jsturtevant cni has been updated, path restriction has been added to workflow and code has been rebased.

vitaliy-leschenko:
Unfortunately github workflow runner doesn't support hyperv isolation.

vitaliy-leschenko:
Still have troubles with os.version

jsturtevant:
For manifest check out https://github.com/kubernetes/kubernetes/blob/9a57dc20a98dcc7a6b2bd1323d4e27f84477cd1b/build/pause/Makefile#L88

kubeadm/flannel/build.ps1
```diff
@@ -0,0 +1,58 @@
...
+                [string] $data = $(docker manifest inspect "$($base):$($tag)" -v)
+                $manifest = $data | ConvertFrom-Json
+                $platform = $manifest.Descriptor.platform
+                $folder = ("docker.io/$($Image):$($flannel)$($suffix)" -replace "/", "_") -replace ":", "-"
+                $img = ("docker.io/$($Image):$($flannel)-$($tag)" -replace "/", "_") -replace ":", "-"
+                Write-Host "Update '~/.docker/manifests/$folder/$img'"
+                $manifest = Get-Content "~/.docker/manifests/$folder/$img" | ConvertFrom-Json
+                $manifest.Descriptor.platform = $platform
+                $manifest | ConvertTo-Json -Depth 10 -Compress | Set-Content "~/.docker/manifests/$folder/$img"
+            }
```
vitaliy-leschenko, line 53: This code is equivalent
https://github.com/kubernetes/kubernetes/blob/9a57dc20a98dcc7a6b2bd1323d4e27f84477cd1b/build/pause/Makefile#L88 
But it failed for now because "~/.docker/manifests" doesn't exists. So, this trick doesn't work.

vitaliy-leschenko: This code doesn't call on pull_request, so check successful, but it is going to fail on merge.
I test it on my test repo https://github.com/vitaliy-leschenko/win-flannel
This code work fine on my linux VM with docker-ce (19.3.12), but failed on github (19.3.13+azure)

vitaliy-leschenko: I found reason why ~/.docker/manifests folder wasn't exist. When we use `azure/docker-login@v1` permissions to folder `~/.docker` corrupts. So, command `docker manifest create` doesn't create manifest files.

kubeadm/flannel/Dockerfile.nanoserver
vitaliy-leschenko: I removed servercore from build

kubeadm/kube-proxy/build.ps1
```diff
@@ -0,0 +1,71 @@
...
+{
+    if ($tag -match "^v(\d+)\.(\d+)\.(\d+)$")
+    {
+        [int]$major = $Matches[1]
+        [int]$minor = $Matches[2]
+        [int]$build = $Matches[3]
+
+        if (($major -gt $minMajor) -or ($major -eq $minMajor -and $minor -gt $minMinor) -or ($major -eq $minMajor -and $minor -eq $minMinor -and $build -ge $minBuild))
+        {
+            buildKubeProxy -tag $tag
```
jsturtevant, line 68: Where we building every version of kube-proxy everytime prior?  How long does this take?

kubeadm/kube-proxy/build.ps1
```diff
@@ -0,0 +1,71 @@
...
+$env:DOCKER_CLI_EXPERIMENTAL = "enabled"
+& docker buildx create --name img-builder --use
+
+$output="docker"
+if ($push.IsPresent) {
+    $output="registry"
+}
+
+
+function buildKubeProxy([string]$tag)
```
jsturtevant, line 19: This looks the same as flannel minus image name can this be made re-usable across flannel and kubeproxy?

jsturtevant (commented):
Thanks for all the hard work here. Left a few comments.

kubeadm/kube-proxy/Dockerfile
```diff
@@ -1,16 +1,22 @@
...
+
+COPY --from=bins /kube-proxy /k/kube-proxy
+COPY --from=bins /utils /utils
 
-RUN mkdir C:\utils; \
-  curl.exe -sLo C:\utils\wins.exe https://github.com/rancher/wins/releases/download/v0.0.4/wins.exe; \
-  curl.exe -sLo C:\utils\yq.exe https://github.com/mikefarah/yq/releases/download/2.4.1/yq_windows_amd64.exe; \
-  "[Environment]::SetEnvironmentVariable('PATH', $env:PATH + ';C:\utils', [EnvironmentVariableTarget]::Machine)"
+ENV PATH="C:\Program Files\PowerShell;C:\utils;C:\Windows\system32;C:\Windows;"
+USER ContainerAdministrator
```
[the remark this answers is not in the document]
vitaliy-leschenko: Unfortunately without it wins doesn't work on nanoserver.

kubeadm/flannel/build.ps1
```diff
@@ -0,0 +1,50 @@
...
+$config = Get-Content .\buildconfig.json | ConvertFrom-Json
+$base = $config.baseimage
+foreach ($flannel in $config.flannel)
+{
+    Write-Host "Build images for flannel version: $flannel"
+
+    $cmd = "docker manifest create $($image):$flannel"
+    foreach($map in $config.tagsMap) 
+    {
+        $cmd = "$cmd --amend $($image):$flannel-$($map.target)"
```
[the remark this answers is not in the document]
vitaliy-leschenko: No, it accumulate all manifests into one manifest list.

kubeadm/flannel/Dockerfile
```diff
@@ -1,30 +1,39 @@
...
+COPY --from=bins /flannel/flanneld.exe /k/flannel/flanneld.exe
+COPY --from=bins /cni /cni
+COPY --from=bins /utils /utils
 
-RUN mkdir C:\utils; \
-  curl.exe -Lo C:\utils\wins.exe https://github.com/rancher/wins/releases/download/v0.0.4/wins.exe; \
-  curl.exe -Lo C:\utils\yq.exe https://github.com/mikefarah/yq/releases/download/2.4.1/yq_windows_amd64.exe; \
-  "[Environment]::SetEnvironmentVariable('PATH', $env:PATH + ';C:\utils', [EnvironmentVariableTarget]::Machine)"
+ENV PATH="C:\Program Files\PowerShell;C:\utils;C:\Windows\system32;C:\Windows;"
+USER ContainerAdministrator
```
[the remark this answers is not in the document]
vitaliy-leschenko: wins.exe doesn't work on ContainerUser. I don't know why, but it doesn't.

.github/workflows/kube-proxy.yml
```diff
@@ -6,12 +6,14 @@ on:
...
-    - uses: azure/docker-login@v1
-      with:
-        username: ${{ secrets.DOCKER_USERNAME }}
-        password: ${{ secrets.DOCKER_SECRET }}
-    - name: Build
-      run: go run kubeadm/hack/publish-kubeproxy.go kubeadm/kube-proxy
+    - name: Build and push images
+      run: |
+        echo ${{ secrets.DOCKER_SECRET }} | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
+        ./build.ps1 -push
```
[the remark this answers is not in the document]
vitaliy-leschenko: Yep. In kubeproxy/build.ps1 we get all k8s versions and build images for all of this.

kubeadm/flannel/Dockerfile
```diff
@@ -1,30 +1,39 @@
...
-ADD hns.psm1 /k/flannel
-COPY --from=builder /gopath/build/setup.exe /k/flannel/setup.exe
+FROM $BASE
 
-RUN mkdir C:\cni; \
-  pushd C:\cni; \
-  curl.exe -Lo cni.tgz https://github.com/containernetworking/plugins/releases/download/v${env:cniVersion}/cni-plugins-windows-amd64-v${env:cniVersion}.tgz; \
-  tar -xf cni.tgz; \
-  rm cni.tgz
+ADD hns.psm1 /k/flannel/hns.psm1
```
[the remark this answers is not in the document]
vitaliy-leschenko: You are right. Good idea. I will update it.

kubeadm/kube-proxy/build.ps1
vitaliy-leschenko: Yes. We are building every version every time.
It takes around 30 minutes.

kubeadm/kube-proxy/build.ps1
vitaliy-leschenko: Maybe it is possible. I don't think about it because this code has some differents.

kubeadm/flannel/build.ps1
```diff
@@ -0,0 +1,50 @@
...
+        & docker buildx build --pull --platform windows/amd64 --output=type=$output -f Dockerfile -t "$($image):$flannel-$($map.target)" --build-arg=BASE="$($base):$($map.source)" --build-arg=flannelVersion=$flannel .
+    }
+
+    if ($push.IsPresent)
+    {
+        Write-Host "Create manifest for $($image):$flannel" -ForegroundColor Yellow
+        Write-Host $cmd
+        Invoke-Expression $cmd
+
+        foreach($map in $config.tagsMap)
```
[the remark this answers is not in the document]
vitaliy-leschenko: I will add some comments into the file.

kubeadm/flannel/Dockerfile
```diff
@@ -1,30 +1,39 @@
...
-SHELL ["powershell", "-NoLogo", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+FROM --platform=linux/amd64 golang as setup
+ENV GOOS=windows
+ENV GOARCH=amd64
+WORKDIR /build
+ADD setup.go .
+RUN go build -o setup.exe setup.go
 
+FROM --platform=linux/amd64 ubuntu as bins
+ARG flannelVersion
```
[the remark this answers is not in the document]
vitaliy-leschenko: No. It required. We need to renew each ARGs after new FROM to use it in the stage.

kubeadm/kube-proxy/build.ps1
vitaliy-leschenko: image
Last run takes 24 minutes (without push to registry)

.github/workflows/flannel.yml
```diff
@@ -0,0 +1,23 @@
+name: flannel-images
+
+on:
+  push:
+    paths:
+      - "kubeadm/flannel/**"
+      - ".github/workflows/flannel.yml"
```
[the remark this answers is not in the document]
vitaliy-leschenko: I combined flannel.yml with test-flannel.yml and kube-proxy.yml with test-kube-proxy.yml

kubeadm/flannel/build.ps1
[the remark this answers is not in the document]
vitaliy-leschenko: https://github.com/kubernetes-sigs/sig-windows-tools/pull/111/commits/310e079ed21d5041b8a0d8c2af847d557c48222c

vitaliy-leschenko:
I rebased my commits to prevent merge

kubeadm/kube-proxy/build.ps1
vitaliy-leschenko: I have done it.

kubeadm/flannel/Dockerfile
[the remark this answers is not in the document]
jsturtevant: I wonder if it is missing something in the container. 

@benmoss @marosset @claudiubelu do you have any ideas on why it might not work?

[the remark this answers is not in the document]
vitaliy-leschenko: `wins cli net get` returns empty string. so `$networkJson = wins cli net get | convertfrom-json` always fails.

fourpastmidnight (changes requested):
Just two quick requests to ensure we don't hard code stone software versions in the dockerfile files for flannel and kube-proxy.

kubeadm/flannel/Dockerfile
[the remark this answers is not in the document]
fourpastmidnight: This happened to me when wins wasn't also installed on the host on which the container was running.

kubeadm/kube-proxy/Dockerfile
```diff
@@ -1,16 +1,22 @@
...
+ARG k8sVersion="v1.19.3"
 
+FROM --platform=linux/amd64 curlimages/curl as bins
 ARG k8sVersion
 
-RUN mkdir -force C:\k\kube-proxy; \
-  pushd C:\k\kube-proxy; \
-  curl.exe -sLO https://dl.k8s.io/${env:k8sVersion}/bin/windows/amd64/kube-proxy.exe
+WORKDIR /utils
+RUN curl -Lo wins.exe https://github.com/rancher/wins/releases/download/v0.0.4/wins.exe
```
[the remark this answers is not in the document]
vitaliy-leschenko: Fixed.

kubeadm/flannel/Dockerfile
```diff
@@ -1,30 +1,39 @@
...
+RUN curl -Lo flanneld.exe https://github.com/coreos/flannel/releases/download/v${flannelVersion}/flanneld.exe
+
+WORKDIR /cni
+RUN curl -Lo cni.tgz https://github.com/containernetworking/plugins/releases/download/v${cniVersion}/cni-plugins-windows-amd64-v${cniVersion}.tgz
+RUN tar -xf cni.tgz
+RUN rm cni.tgz
+
+WORKDIR /utils
+RUN curl -Lo wins.exe https://github.com/rancher/wins/releases/download/v0.0.4/wins.exe
+RUN curl -Lo yq.exe https://github.com/mikefarah/yq/releases/download/2.4.1/yq_windows_amd64.exe
```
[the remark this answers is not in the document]
vitaliy-leschenko: Fixed.

kubeadm/flannel/Dockerfile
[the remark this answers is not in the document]
fourpastmidnight: Hmm, on second thought, that's not the same issue as I experienced. Wins can't even access the named pipe. Sorry.

kubeadm/flannel/Dockerfile
```diff
@@ -1,30 +1,43 @@
...
 
+FROM --platform=linux/amd64 curlimages/curl as bins
 ARG cniVersion
+ARG flannelVersion
+ARG winsVersion
+ARG jqVersion
+
+WORKDIR /utils
+RUN curl -Lo wins.exe https://github.com/rancher/wins/releases/download/v${winsVersion}/wins.exe
+RUN curl -Lo yq.exe https://github.com/mikefarah/yq/releases/download/${jqVersion}/yq_windows_amd64.exe
```
fourpastmidnight, line 22: Shouldn't the command line switches be `-LO`?

fourpastmidnight (commented):
Noticed possibly incorrect switches to curl.

kubeadm/flannel/Dockerfile
vitaliy-leschenko: No. `-Lo` is correct. `-LO` is a little bit different.

fourpastmidnight: OH! You're right. I got `o` and `O` mixed up, thinking `O` was the just download the file and save as the same name as the leaf of the url--but that is `o`.

kubeadm/flannel/Dockerfile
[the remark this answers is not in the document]
jsturtevant: I am concerned on moving moving to nano server and switching to ContainerAdministrator as it is a big change for the way the image is used.  If we aren't able to track this down, we should go back to servercore so we can merge this and open a separate issue to track the change to nano without use of ContainerAdministrator.  Does that sound good?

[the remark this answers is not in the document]
vitaliy-leschenko: Ok. I will switch back to servercore.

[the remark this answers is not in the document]
vitaliy-leschenko: But I think that use ContainerAdministrator is not a big risk because we already use wins.exe and run code at host on SYSTEM account.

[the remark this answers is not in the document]
vitaliy-leschenko: @jsturtevant What do you think about build both: servercore (will work on User) and nanoserver (will work on Admin).
So we won't break current flow and will add possibility use nano (maybe with more risks)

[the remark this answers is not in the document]
vitaliy-leschenko: Rollback to servercore also affects build time.
flannel: from 3m (nanoserver) to 36m (servercore).
kube-proxy: from 25m (nanoserver) to 2h+ (servercore).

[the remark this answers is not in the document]
vitaliy-leschenko: I switched to nanoserver again because:
- we can't build kube-proxy for all k8s version for all windows version (3h is not enough for it)
- we run flannel.exe and kubeproxy.exe via wins as SYSTEM, so ContainerUser or ContainerAdministrator doesn't matter.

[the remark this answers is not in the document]
vitaliy-leschenko: @jsturtevant What do you think about using on nanoserver+ContainerAdmin in case when we can't automate build based on servercore? 
Do you think that using ContainerAdminstrator is unacceptable?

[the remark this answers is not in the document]
jsturtevant: I get your points though I still don't feel good about making that change.  @marosset do you have any thoughts or ideas on why wins.exe doesn't work in nanoserver?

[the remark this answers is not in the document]
vitaliy-leschenko: @marosset @jsturtevant a bit details: Why nanoserver doesn't support pipe mounting on ContainerUser.
wins.exe doesn't work because it has no access to \\.\pipe\rancher_wins when starts on ContainerUser.
It works only on ContainerAdministrator.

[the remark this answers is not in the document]
vitaliy-leschenko: @jsturtevant I found answers.
Default user for mcr.microsoft.com/windows/servercore is ContainerAdministrator
Default user for mcr.microsoft.com/windows/nanoserver is ContainerUser.

Proof:
image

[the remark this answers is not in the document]
vitaliy-leschenko: And final message. I tried to build test image based on servercore
```
FROM mcr.microsoft.com/windows/servercore:1809
ADD https://github.com/rancher/wins/releases/download/v0.0.4/wins.exe /wins.exe
USER ContainerUser
```
Then I run it
```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: wins-test
  labels:
    tier: node
    app: wins-test
spec:
  selector:
    matchLabels:
      app: wins-test
  template:
    metadata:
      labels:
        tier: node
        app: wins-test
    spec:
      hostNetwork: true
      nodeSelector:
        kubernetes.io/os: 'windows'
      containers:
      - name: wins
        image: vleschenko/wins-test:user
        command:
        - ping 
        args:
        - -t
        - localhost
        volumeMounts:
        - name: wins
          mountPath: \\.\pipe\rancher_wins
        env:
        - name: POD_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.name
        - name: POD_NAMESPACE
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
      volumes:
      - name: wins
        hostPath:
          path: \\.\pipe\rancher_wins
          type: null
```
When I connected to pod and tried run `wins cli net get` wins failed with the same error as on nanoserver - Access is denied.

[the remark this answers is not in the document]
jsturtevant: Thanks for digging into this!  Learned something new today :-).  I'll give it a final review today.

vitaliy-leschenko:
commit has been rebased

vitaliy-leschenko:
@jsturtevant do you have any other comments for the PR? You promised to final review a week ago.

kubeadm/buildx.psm1
```diff
@@ -0,0 +1,58 @@
...
+    $command = "docker buildx build --platform windows/amd64 --output=type=$output -f Dockerfile -t $name"
+    foreach($arg in $args)
+    {
+        $command = "$command --build-arg=$arg"
+    }
+    $command = "$command ."
+    Invoke-Expression $command
+}
+
+function Add-Manifest([string]$name, [string[]]$items, [string[]]$bases)
```
jsturtevant, line 29: This looks to push the manifest as well.  Could you rename it "Push-Manifest" to reflect that?

kubeadm/flannel/build.ps1
```diff
@@ -0,0 +1,36 @@
...
+        $items += $current
+        New-Build -name $current -output $output -args @("BASE=$($base):$($map.source)", "flannelVersion=$flannel")
+    }
+
+    if ($push.IsPresent)
+    {
+        Add-Manifest -name "$($image):$flannel" -items $items -bases $bases
+    }
+
+    Write-Host
```
jsturtevant, line 35: is this an extra write host?

kubeadm/flannel/build.ps1
```diff
@@ -0,0 +1,36 @@
...
+    Write-Host "Build images for flannel version: $flannel"
+
+    [string[]]$items = @()
+    [string[]]$bases = @()
+    foreach($map in $config.tagsMap) 
+    {
+        $bases += "$($base):$($map.source)"
+        $current = "$($image):$flannel-$($map.target)"
+        $items += $current
+        New-Build -name $current -output $output -args @("BASE=$($base):$($map.source)", "flannelVersion=$flannel")
```
jsturtevant, line 27: `map.source` could be more clear here like `sourceTag`? or Maybe `map` could be named `tag`?

kubeadm/buildx.psm1
```diff
@@ -0,0 +1,58 @@
...
+        [string[]]$args
+    )
+
+    $command = "docker buildx build --platform windows/amd64 --output=type=$output -f Dockerfile -t $name"
+    foreach($arg in $args)
+    {
+        $command = "$command --build-arg=$arg"
+    }
+    $command = "$command ."
+    Invoke-Expression $command
```
jsturtevant, line 26: could we log the command and the tag we are building?  Looking through the logs it is hard to figure out what tags are being built.

jsturtevant (commented):
It looks like we lost the ability to tag the versions per build somewhere in transition to buildx?

Something like:
sigwindowstools/flannel:v0.12 (multi arch manifest that points to the latest of the following images)
sigwindowstools/flannel:v0.12-2019-{git_tag}

If we are doing this we may not want to build the images on master but instead on a push of [git tag instead of push to master](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onevent_nametypes) so we can control the changes more.  

```yaml
on:
  push:
    tags:        
      - v*
```

kubeadm/buildx.psm1
vitaliy-leschenko: Ok. I did it.

kubeadm/flannel/build.ps1
vitaliy-leschenko: Extra Write-Host has been removed.

kubeadm/flannel/build.ps1
vitaliy-leschenko: I renamed $map to $tag

kubeadm/buildx.psm1
vitaliy-leschenko: I added `Write-Host $command` before each `Invoke-Expression $command`

vitaliy-leschenko:
@jsturtevant I made changes based on your comments. Please review it.

kubeadm/flannel/Dockerfile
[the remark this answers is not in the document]
marosset: Very interesting.

kubeadm/kube-proxy/build.ps1
```diff
@@ -0,0 +1,53 @@
...
+$versions = (curl -L k8s.gcr.io/v2/kube-proxy/tags/list | ConvertFrom-Json).tags
+foreach($version in $versions)
+{
+    if ($version -match "^v(\d+)\.(\d+)\.(\d+)$")
+    {
+        [int]$major = $Matches[1]
+        [int]$minor = $Matches[2]
+        [int]$build = $Matches[3]
+
+        if (($major -gt $minMajor) -or ($major -eq $minMajor -and $minor -gt $minMinor) -or ($major -eq $minMajor -and $minor -eq $minMinor -and $build -ge $minBuild))
```
marosset, line 48: This logic can probably be simplified by using [version] accelerators in powershell 
https://invoke-thebrain.com/2018/12/comparing-version-numbers-powershell/

marosset, line 48: Can we log which versions of kube-proxy we build for and which we don't?

marosset:
I added a few comments in the script to build various kube-proxy images to help simplify the logic and debugging.
Other than this PR looks good to me.
I'll let @jsturtevant give the final LGTM since he's been much more active in the review

kubeadm/kube-proxy/build.ps1
vitaliy-leschenko: I added some logs when we build new image and why we don't.

vitaliy-leschenko: I updated this code to use [version]

vitaliy-leschenko:
@jsturtevant @marosset I updated code by your comments.

jsturtevant:
@vitaliy-leschenko have you tried using the push since moving to buildx?  

I ran `pwsh build.ps1 -image jsturtevant/flannel-tools -push` and it is failing with:

```
 => ERROR exporting to image                                                                                                                                                                     1.2s
 => => exporting layers                                                                                                                                                                          0.0s
 => => exporting manifest sha256:8f346e15ccbaf34634fe53b8713979efce2ad344b10a036f6fab4195b978f53e                                                                                                0.0s
 => => exporting config sha256:45552700afa92c2177dd351169d71b5cf207570b48c26c9e4d5f20d4ee4c39b8                                                                                                  0.0s
 => => pushing layers                                                                                                                                                                            0.8s
 => => pushing manifest for docker.io/jsturtevant/flannel-tools:v0.12.0-2004                                                                                                                     0.2s
------
 > exporting to image:
------
failed to solve: rpc error: code = Unknown desc = failed commit on ref "manifest-sha256:8f346e15ccbaf34634fe53b8713979efce2ad344b10a036f6fab4195b978f53e": unexpected status: 500 Internal Server Error
```

vitaliy-leschenko:
@jsturtevant Yes, I have. But something strange is happening:
1. when I run it on my local VM it finish successful.
2. when I try to run it on gitlab build server it fails.
3. when I try to re-run previously successful jobs on my test repo (https://github.com/vitaliy-leschenko/k8s-docker-images) then fail.

I don't understand.

vitaliy-leschenko:
I cleanup my local buildx (and remove image moby/buildkit:buildx-stable-1) and now I have the same error.
So, this is a new buildkit issue but it block us.

vitaliy-leschenko:
@jsturtevant I found reason why build was failed:
A few days ago was released new `moby/buildkit` and it has a bug.
I tried to use previous one and it is OK.
So, I fix buildkit version via `docker buildx create --name img-builder --use --driver docker-container --driver-opt image=moby/buildkit:v0.7.2` in `buildx.psm1`

Build (and push) has been fixed.

kubeadm/buildx.psm1
```diff
@@ -0,0 +1,60 @@
...
+        $command = "$command --build-arg=$arg"
+    }
+    $command = "$command ."
+    Write-Host $command
+    Invoke-Expression $command
+}
+
+function Push-Manifest([string]$name, [string[]]$items, [string[]]$bases)
+{
+    $command = "docker manifest create $name";
```
jsturtevant, line 32: during my local testing if this is called multiple times then it will continually append the manifest lists. This is not an issue in CI as it is not persisted but if someone wanted to use this to push to custom repo they could run into it.  

If an existing manifest is present on the machine of the same name can we clear it?  This should either print a warning or maybe even take a flag?

kubeadm/buildx.psm1
```diff
@@ -0,0 +1,60 @@
...
+    foreach($item in $items)
+    {
+        $command = "$command --amend $item"
+    }
+    Write-Host $command
+    Invoke-Expression $command
+
+    for ($i = 0; $i -lt $items.Length; $i++) {
```
jsturtevant, line 40: Can we just add a note this is work around until the next major release for docker cli in 20.x : https://github.com/docker/cli/pull/2578

jsturtevant (commented):
Awesome work. I noticed one issue when run multiple times in same env.  Otherwise looks great.  Thanks for all the work here!

kubeadm/buildx.psm1
vitaliy-leschenko: Yes. I did it.

kubeadm/buildx.psm1
vitaliy-leschenko: I didn't think about it. Ok. I will try to remove previous manifest if it exists and warning about it.

vitaliy-leschenko: Check exists manifest has been done.

jsturtevant:
I am doing some final testing against the images in a cluster.  Otherwise this looks good

kubeadm/flannel/flannel-host-gw.yml
```diff
@@ -179,9 +179,9 @@ spec:
         effect: NoSchedule
       containers:
       - name: kube-flannel
-        image: sigwindowstools/flannel:0.12.0
+        image: sigwindowstools/flannel:v0.13.0
```
jsturtevant, line 182: This added a `v` in the image. It would make it more consistent but would break the pattern of the older images

jsturtevant:
Since we moved to nano-server, the yaml files for Flannel and kube-proxy where also switched to `pwsh` otherwise the images boot with `hcsshim::System::CreateProcess: failure in a Windows system call: The system cannot find the file specified. (0x2) extra info: {"CommandLine":"powershell -file /etc/kube-flannel-windows/run.ps1`


This is a breaking change for anyone that might be using the images. Since in the past we don't have tags for each release, the "latest" image (in this case the  `kube-proxy:v1.20.0` and `flannel:0.12.0`) if pulled fresh would break folks.  We may want to tag the current versions so folks have a way to grab those specifically.  Thoughts? @marosset as well

kubeadm/flannel/flannel-host-gw.yml
vitaliy-leschenko: I know. I added `v` because `quay.io/coreos/flannel` has it.

vitaliy-leschenko: Do you think it was bad idea?

vitaliy-leschenko:
To avoid backward compatibility issues with exists images I propose use `-nanoserver` suffix.

For example: 
- `sigwindowstools/kube-proxy:v1.20.0-nanoserver`
- `sigwindowstools/flannel:v0.13.0-nanoserver`

So, any exists images won't be affected.

vitaliy-leschenko:
@jsturtevant @marosset what do you think about `-nanoserver` suffix?

kubeadm/flannel/flannel-host-gw.yml
jsturtevant: I think I've been burned but the missing `v` more than once but wonder about consistency, If we use end up going with the `nanoserver` suffix this wouldn't be to much of an issue.

jsturtevant:
That's probably the easiest approach, We can update the yaml's to point to those.

vitaliy-leschenko:
Ok. I will do it now.

jsturtevant:
/lgtm
/approve

@vitaliy-leschenko Thanks for all your hard work and persistence on this one!  :tada: :100:

vitaliy-leschenko:
@jsturtevant please create tag to trigger flannel building.

jsturtevant:
Thanks! https://github.com/kubernetes-sigs/sig-windows-tools/releases/tag/v0.1.5

vitaliy-leschenko:
@jsturtevant we have a two problems:
1. password to docker hub must be changed as soon as possible.
2. we need to escape password in workflow yaml

vitaliy-leschenko:
#134 can fix it

PR #112, opened 2020-10-26
Adding containerd support to kubeadm related scripts/specs/etc

marosset:
Adding support for configuring a cluster that uses containerd to the existing kubeadm implementionstations.

Verify the following:
- [x] Add docker node to a cluster using overlay networking
- [x] Add containerd node to a cluster using overlay networking
- [x] Add docker node to a cluster using bridge/host-gw networking
- [x] Add containerd node to a cluster using bridge/host-gw networking

Notes:
Install-containerd.ps1 needs to be downloaded and run before running PrepareNode.ps1.
This assumes a completely vanilla Windows installation and may need to enable Windows features which will require a reboot.

After running Install-Containerd.ps1 to completion run PrepaseNode.ps1 -ContainerRuntime containerd.

Kube-proxy/CNI deployment should be the same as with docker.

All of these will be captured in updates to kubernetes/website updates to https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/adding-windows-nodes/

### DIFF
--- a/kubeadm/flannel/flannel-host-gw.yml
+++ b/kubeadm/flannel/flannel-host-gw.yml
@@ -17,15 +17,34 @@ data:
     mkdir -force /host/k/flannel
     mkdir -force /host/k/flannel/var/run/secrets/kubernetes.io/serviceaccount
 
-    $cniJson = get-content /etc/kube-flannel-windows/cni-conf.json | ConvertFrom-Json
+    $containerRuntime = "docker"
+    if (Test-Path /host/etc/cni/net.d/0-containerd-nat.json) {
+      $containerRuntime = "containerd"
+    }
+
+    Write-Host "Configuring CNI for $containerRuntime"
+
     $serviceSubnet = yq r /etc/kubeadm-config/ClusterConfiguration networking.serviceSubnet
     $podSubnet = yq r /etc/kubeadm-config/ClusterConfiguration networking.podSubnet
     $networkJson = wins cli net get | convertfrom-json
 
-    $cniJson.delegate.policies[0].Value.ExceptionList = $serviceSubnet, $podSubnet, $networkJson.SubnetCIDR
-    $cniJson.delegate.policies[1].Value.DestinationPrefix = $serviceSubnet
-    $cniJson.delegate.policies[2].Value.DestinationPrefix = $networkJson.AddressCIDR
-    Set-Content -Path /host/etc/cni/net.d/10-flannel.conf ($cniJson | ConvertTo-Json -depth 100)
+    if ($containerRuntime -eq "docker") {
+      $cniJson = get-content /etc/kube-flannel-windows/cni-conf.json | ConvertFrom-Json
+
+      $cniJson.delegate.policies[0].Value.ExceptionList = $serviceSubnet, $podSubnet, $networkJson.SubnetCIDR
+      $cniJson.delegate.policies[1].Value.DestinationPrefix = $serviceSubnet
+      $cniJson.delegate.policies[2].Value.DestinationPrefix = $networkJson.AddressCIDR
+
+      Set-Content -Path /host/etc/cni/net.d/10-flannel.conf ($cniJson | ConvertTo-Json -depth 100)
+    } elseif ($containerRuntime -eq "containerd") {
+      $cniJson = get-content /etc/kube-flannel-windows/cni-conf-containerd.json | ConvertFrom-Json
+
+      $cniJson.delegate.AdditionalArgs[0].Value.Settings.Exceptions = $serviceSubnet, $podSubnet, $networkJson.SubnetCIDR
+      $cniJson.delegate.AdditionalArgs[1].Value.Settings.DestinationPrefix = $serviceSubnet
+      $cniJson.delegate.AdditionalArgs[2].Value.Settings.DestinationPrefix = $networkJson.AddressCIDR
+
+      Set-Content -Path /host/etc/cni/net.d/10-flannel.conf ($cniJson | ConvertTo-Json -depth 100)
+    }
 
     cp -force /etc/kube-flannel/net-conf.json /host/etc/kube-flannel
     cp -force -recurse /cni/* /host/opt/cni/bin
@@ -69,6 +88,53 @@ data:
               "Type": "ROUTE",
               "DestinationPrefix": "",
               "NeedEncap": true
+            }
+          }
+        ]
+      }
+    }
+  cni-conf-containerd.json: |
+    {
+      "cniVersion": "0.2.0",
+      "name": "cbr0",
+      "type": "flannel",
+      "capabilities": {
+        "portMappings": true,
+        "dns": true
+      },
+      "delegate": {
+        "type": "sdnbridge",
+        "optionalFlags" : {
+          "forceBridgeGateway" : true
+        },
+        "AdditionalArgs": [
+          {
+            "Name": "EndpointPolicy",
+            "Value": {
+              "Type": "OutBoundNAT",
+              "Settings": {
+                "Exceptions": []
+              }
+            }
+          },
+          {
+            "Name": "EndpointPolicy",
+            "Value": {
+              "Type": "SDNROUTE",
+              "Settings": {
+                "DestinationPrefix": "",
+                "NeedEncap": true
+              }
+            }
+          },
+          {
+            "Name": "EndpointPolicy",
+            "Value": {
+              "Type": "SDNROUTE",
+              "Settings": {
+                "DestinationPrefix": "",
+                "NeedEncap": true
+              }
             }
           }
         ]
--- a/kubeadm/flannel/flannel-overlay.yml
+++ b/kubeadm/flannel/flannel-overlay.yml
@@ -17,20 +17,40 @@ data:
     mkdir -force /host/k/flannel
     mkdir -force /host/k/flannel/var/run/secrets/kubernetes.io/serviceaccount
 
-    $cniJson = get-content /etc/kube-flannel-windows/cni-conf.json | ConvertFrom-Json
+    $containerRuntime = "docker"
+    if (Test-Path /host/etc/cni/net.d/0-containerd-nat.json) {
+      $containerRuntime = "containerd"
+    }
+
+    Write-Host "Configuring CNI for $containerRuntime"
+
     $serviceSubnet = yq r /etc/kubeadm-config/ClusterConfiguration networking.serviceSubnet
     $podSubnet = yq r /etc/kubeadm-config/ClusterConfiguration networking.podSubnet
     $networkJson = wins cli net get | convertfrom-json
 
-    $cniJson.delegate.policies[0].Value.ExceptionList = $serviceSubnet, $podSubnet
-    $cniJson.delegate.policies[1].Value.DestinationPrefix = $serviceSubnet
-    Set-Content -Path /host/etc/cni/net.d/10-flannel.conf ($cniJson | ConvertTo-Json -depth 100)
+    if ($containerRuntime -eq "docker") {
+      $cniJson = get-content /etc/kube-flannel-windows/cni-conf.json | ConvertFrom-Json
+
+      $cniJson.delegate.policies[0].Value.ExceptionList = $serviceSubnet, $podSubnet
+      $cniJson.delegate.policies[1].Value.DestinationPrefix = $serviceSubnet
+
+      Set-Content -Path /host/etc/cni/net.d/10-flannel.conf ($cniJson | ConvertTo-Json -depth 100)
+    } elseif ($containerRuntime -eq "containerd") {
+      $cniJson = get-content /etc/kube-flannel-windows/cni-conf-containerd.json | ConvertFrom-Json
+
+      $cniJson.delegate.AdditionalArgs[0].Value.Settings.Exceptions = $serviceSubnet, $podSubnet
+      $cniJson.delegate.AdditionalArgs[1].Value.Settings.DestinationPrefix = $serviceSubnet
+      $cniJson.delegate.AdditionalArgs[2].Value.Settings.ProviderAddress = $networkJson.AddressCIDR.Split('/')[0]
+
+      Set-Content -Path /host/etc/cni/net.d/10-flannel.conf ($cniJson | ConvertTo-Json -depth 100)
+    }
 
     cp -force /etc/kube-flannel/net-conf.json /host/etc/kube-flannel
     cp -force -recurse /cni/* /host/opt/cni/bin
     cp -force /k/flannel/* /host/k/flannel/
     cp -force /kube-proxy/kubeconfig.conf /host/k/flannel/kubeconfig.yml
     cp -force /var/run/secrets/kubernetes.io/serviceaccount/* /host/k/flannel/var/run/secrets/kubernetes.io/serviceaccount/
+
     wins cli process run --path /k/flannel/setup.exe --args "--mode=overlay --interface=Ethernet"
     wins cli route add --addresses 169.254.169.254
     wins cli process run --path /k/flannel/flanneld.exe --args "--kube-subnet-mgr --kubeconfig-file /k/flannel/kubeconfig.yml" --envs "POD_NAME=$env:POD_NAME POD_NAMESPACE=$env:POD_NAMESPACE"
@@ -58,6 +78,49 @@ data:
               "Type": "ROUTE",
               "DestinationPrefix": "",
               "NeedEncap": true
+            }
+          }
+        ]
+      }
+    }
+  cni-conf-containerd.json: |
+    {
+      "name": "flannel.4096",
+      "cniVersion": "0.2.0",
+      "type": "flannel",
+      "capabilities": {
+        "portMappings": true,
+        "dns": true
+      },
+      "delegate": {
+        "type": "sdnoverlay",
+        "AdditionalArgs": [
+          {
+            "Name": "EndpointPolicy",
+            "Value": {
+              "Type": "OutBoundNAT",
+              "Settings" : {
+                "Exceptions": []
+              }
+            }
+          },
+          {
+            "Name": "EndpointPolicy",
+            "Value": {
+              "Type": "SDNROUTE",
+              "Settings": {
+                "DestinationPrefix": "",
+                "NeedEncap": true
+              }
+            }
+          },
+          {
+            "Name":"EndpointPolicy",
+            "Value":{
+              "Type":"ProviderAddress",
+                "Settings":{
+                    "ProviderAddress":""
+              }
             }
           }
         ]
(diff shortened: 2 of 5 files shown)


michmike (approved):
/lgtm
/approve

kubeadm/scripts/Install-Containerd.ps1
```diff
@@ -0,0 +1,153 @@
...
+}
+"@ | Set-Content "c:\etc\cni\net.d\0-containerd-nat.json" -Force
+
+Write-Output "Registering ContainerD as a service"
+containerd.exe --register-service
+
+Write-Output "Starting ContainerD service"
+Start-Service containerd
+
+Write-Ouput "Done - please remember to add '--cri-socket `"npipe:////./pipe/containerd-containerd`"' to your kubeadm join command"
```
vitaliy-leschenko, line 153: Typos has been found there. `Write-Output` instead of `Write-Ouput`

marosset: Fixed!

marosset:
/assign @benmoss @jsturtevant 

I think this is ready for review.

marosset:
I've started some doc updates for using these changes https://github.com/kubernetes/website/pull/24862

Related to https://github.com/kubernetes/enhancements/issues/1001

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -11,14 +11,20 @@ This script assists with joining a Windows node to a cluster.
 .PARAMETER KubernetesVersion
 Kubernetes version to download and use
 
+.PARAMETER ContainerRuntime
+Container that Kubernetes will use. (Docker or containerD)
+
 .EXAMPLE
-PS> .\PrepareNode.ps1 -KubernetesVersion v1.17.0
+PS> .\PrepareNode.ps1 -KubernetesVersion v1.19.1 -ContainerRuntime containerD
```
jsturtevant, line 18: This move fast 1.19.3 is out already :smile:

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -11,14 +11,20 @@ This script assists with joining a Windows node to a cluster.
...
 
 #>
 
 Param(
     [parameter(Mandatory = $true, HelpMessage="Kubernetes version to use")]
-    [string] $KubernetesVersion
+    [string] $KubernetesVersion,
+    [parameter(HelpMessage="Container runtime that Kubernets will use")]
+    [ValidateSet("containerD", "Docker")]
+    [string] $ContainerRuntime = "Docker"
```
jsturtevant, line 27: with Containerd going Stable in 1.20 should the default be ContainerD?

kubeadm/scripts/Install-Containerd.ps1
```diff
@@ -0,0 +1,158 @@
...
+mkdir -Force c:\etc\cni\net.d | Out-Null
+
+Write-Output "Getting SDN CNI binaries"
+DownloadFile "c:\opt\cni\cni-plugins.zip" https://github.com/microsoft/windows-container-networking/releases/download/v0.2.0/windows-container-networking-cni-amd64-v0.2.0.zip
+Expand-Archive -Path "c:\opt\cni\cni-plugins.zip" -DestinationPath "c:\opt\cni\bin" -Force
+
+mkdir -Force c:\k | Out-Null
+DownloadFile "c:\k\hns.psm1" https://github.com/Microsoft/SDN/raw/master/Kubernetes/windows/hns.psm1
+Import-Module "c:\k\hns.psm1"
+New-HnsNetwork -Type NAT -Name nat
```
jsturtevant, line 123: Is `New-HnsNetwork -Type NAT -Name nat` the same as `docker network create -d nat host` in prepare-node? It looks like the nat network is configured in both scripts in different ways.  

Is there a reason the network config for the nat needs to be done here and not in prepare node?

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -51,9 +57,13 @@ DownloadFile $kubeletBinPath https://dl.k8s.io/$KubernetesVersion/bin/windows/am
 DownloadFile "$global:KubernetesPath\kubeadm.exe" https://dl.k8s.io/$KubernetesVersion/bin/windows/amd64/kubeadm.exe
 DownloadFile "$global:KubernetesPath\wins.exe" https://github.com/rancher/wins/releases/download/v0.0.4/wins.exe
 
-# Create host network to allow kubelet to schedule hostNetwork pods
-Write-Host "Creating Docker host network"
-docker network create -d nat host
+if ($ContainerRuntime -eq "Docker") {
```
jsturtevant, line 60: There should probably be a check that Containerd/docker is installed on the machine.  Could this script call Install-containerd?

kubeadm/scripts/Install-Containerd.ps1
```diff
@@ -0,0 +1,158 @@
...
+    Write-Host("Downloading $source to $destination")
+    curl.exe --silent --fail -Lo $destination $source
+
+    if (!$?) {
+        Write-Error "Download $source failed"
+        exit 1
+    }
+}
+
+function CalculateSubNet {
```
jsturtevant, line 43: This is a complex calculation.  Do we have an example of what it does in docs here?  Maybe add a test?

kubeadm/flannel/flannel-overlay.yml
```diff
@@ -17,20 +17,37 @@ data:
...
     $networkJson = wins cli net get | convertfrom-json
 
-    $cniJson.delegate.policies[0].Value.ExceptionList = $serviceSubnet, $podSubnet
-    $cniJson.delegate.policies[1].Value.DestinationPrefix = $serviceSubnet
-    Set-Content -Path /host/etc/cni/net.d/10-flannel.conf ($cniJson | ConvertTo-Json -depth 100)
+    if ($containerRuntime -eq "docker") {
+      $cniJson = get-content /etc/kube-flannel-windows/cni-conf.json | ConvertFrom-Json
+
+      $cniJson.delegate.policies[0].Value.ExceptionList = $serviceSubnet, $podSubnet
+      $cniJson.delegate.policies[1].Value.DestinationPrefix = $serviceSubnet
```
jsturtevant, line 33: Why isn't the additional `$cniJson.delegate.policies[2].Value.DestinationPrefix = $networkJson.AddressCIDR` needed for overlay?

kubeadm/flannel/flannel-host-gw.yml
```diff
@@ -17,15 +17,32 @@ data:
     mkdir -force /host/k/flannel
     mkdir -force /host/k/flannel/var/run/secrets/kubernetes.io/serviceaccount
 
-    $cniJson = get-content /etc/kube-flannel-windows/cni-conf.json | ConvertFrom-Json
+    $containerRuntime = "docker"
+    if (Test-Path /host/etc/cni/net.d/0-containerd-nat.json) {
+      $containerRuntime = "containerd"
```
jsturtevant, line 22: Can we add a logging statement that outputs what runtime is selected?

kubeadm/flannel/flannel-overlay.yml
```diff
@@ -17,20 +17,37 @@ data:
     mkdir -force /host/k/flannel
     mkdir -force /host/k/flannel/var/run/secrets/kubernetes.io/serviceaccount
 
-    $cniJson = get-content /etc/kube-flannel-windows/cni-conf.json | ConvertFrom-Json
+    $containerRuntime = "docker"
+    if (Test-Path /host/etc/cni/net.d/0-containerd-nat.json) {
+      $containerRuntime = "containerd"
```
jsturtevant, line 22: Same here for the logging statement

jsturtevant (commented):
Left a few comments.

kubeadm/scripts/Install-Containerd.ps1
marosset: They should be effectively the same.
Let me move this to PrepareNode.ps1

kubeadm/scripts/PrepareNode.ps1
marosset: Since we haven't published a lot of docs previously on how to configure and run with containerD I thought we can keep the defaults for these scripts using Docker for 1 more release.
What do you think?

kubeadm/scripts/PrepareNode.ps1
marosset: Since Install-Containerd.ps1 may need to be run mutliple times I'd rather not call it.
We can check if the service is installed tho.

kubeadm/flannel/flannel-overlay.yml
marosset: I'm not sure but I haven't seen that used in any other examples when configuring an overlay network.
Ex: https://github.com/microsoft/SDN/blob/3e54e29aa41bf751bc8a5f528a466b285b4b0291/Kubernetes/windows/helper.v2.psm1#L789-L810

kubeadm/scripts/Install-Containerd.ps1
marosset: Let me add an example.
I can follow up with a test but since we don't have any other pester/powershell tests currently I think we can just add all that later.

kubeadm/flannel/flannel-overlay.yml
ksubrmnn: In Overlay, local pods talk to local host via HostRoute (which is a policy applied on the network itself by Flannel, developed specifically for Overlay). Only L2Bridge needs the SDN Route to management IP. Not sure on the specifics of how it works though

kubeadm/scripts/PrepareNode.ps1
ksubrmnn: +1 for keeping docker for one more release. We can switch to containerd as default once people have used it more imo

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -32,6 +38,18 @@ function DownloadFile($destination, $source) {
     }
 }
 
+if ($ContainerRuntime -eq "Docker") {
```
ksubrmnn, line 41: Kubeadm uses named pipe detection to check for these services: https://github.com/kubernetes/kubernetes/blob/b6b01e17ec8149d53fa9ab5812f9c5947e69a987/cmd/kubeadm/app/util/runtime/runtime_windows.go

May be worth making it consistent? I don't feel strongly about it.

ksubrmnn (changes requested):

kubeadm/scripts/PrepareNode.ps1
ksubrmnn: Actually I think we should keep it consistent on second thought.

kubeadm/scripts/PrepareNode.ps1
jsturtevant: sgtm!

kubeadm/flannel/flannel-overlay.yml
jsturtevant: Thanks for the clarity.  Should we add a comment here for others in the future?

kubeadm/scripts/PrepareNode.ps1
jsturtevant: Why would Install-Containerd.ps1 need to be called more than once?  

Checking for the service makes sense.

marosset: If the required windows features are not enabled the script enables them and then prompts you to reboot and re-run the script.

kubeadm/scripts/PrepareNode.ps1
marosset: Updated

kubeadm/flannel/flannel-overlay.yml
marosset: I think we should document that but i'm not sure this inline'd powershell script in a deployment file is the appropriate place.

kubeadm/flannel/flannel-overlay.yml
```diff
@@ -63,6 +82,40 @@ data:
...
+          {
+            "Name": "EndpointPolicy",
+            "Value": {
+              "Type": "SDNROUTE",
+              "Settings": {
+                "DestinationPrefix": "",
+                "NeedEncap": true
+              }
+            }
+          }
```
daschott, line 115: Should there be PA address specified here as well?

See https://github.com/microsoft/windows-container-networking/blob/master/example/flannel_overlay.conf

kubeadm/flannel/flannel-host-gw.yml
```diff
@@ -74,6 +93,50 @@ data:
...
+    {
+      "cniVersion": "0.2.0",
+      "name": "cbr0",
+      "type": "flannel",
+      "capabilities": {
+        "portMappings": true,
+        "dns": true
+      },
+      "delegate": {
+        "type": "sdnbridge",
```
daschott, line 106: @nagiesek
Should we be setting the forceBridgeGateway flag always since this is used by Flannel? I believe otherwise the .2 address which is reserved could be attempted to be assigned to pods?
```
        "optionalFlags" : {
            "forceBridgeGateway" : true
        }
```
Source: https://github.com/microsoft/windows-container-networking/blob/master/example/flannel_l2bridge.conf

I think similar may also be needed in the overlay CNI config?

kubeadm/flannel/flannel-host-gw.yml
```diff
@@ -74,6 +93,50 @@ data:
         ]
       }
     }
+  cni-conf-containerd.json: |
+    {
+      "cniVersion": "0.2.0",
+      "name": "cbr0",
+      "type": "flannel",
```
daschott, line 100: @nagiesek
win-bridge and win-overlay support "loopbackDSR" 
https://github.com/containernetworking/plugins/tree/master/plugins/main/windows/win-bridge

- is this also implemented in the SDN plugins and should be added if DSR is enabled?

daschott (commented):
I left a few notes/open questions on the CNI configs

kubeadm/flannel/flannel-overlay.yml
marosset: I can add it if you think it would be of benefit.

kubeadm/flannel/flannel-host-gw.yml
marosset: This implementation relies on wins and I'm not sure if there is a great way to check if DSR is enabled.
Let me make an another issue in this repo for DSR enablement and we can merge that support with a seperate PR.

kubeadm/flannel/flannel-overlay.yml
```diff
@@ -63,6 +83,49 @@ data:
...
+                "NeedEncap": true
+              }
+            }
+          },
+          {
+            "Name":"EndpointPolicy",
+            "Value":{
+              "Type":"ProviderAddress",
+                "Settings":{
+                    "ProviderAddress":""
```
jsturtevant, line 122: @daschott  Does this need to be set to the ip of the host?

https://github.com/microsoft/windows-container-networking/blob/a1b979e8f5cbccd25719747ba20c76ca8d1ae156/example/flannel_overlay.conf#L37-L39

What does it do?

jsturtevant:
Over all looks good. Had one question on the PA address section

kubeadm/flannel/flannel-overlay.yml
jsturtevant: I see this is set to the host ip on https://github.com/kubernetes-sigs/sig-windows-tools/pull/112/files/dac8d9cae606b03484fd294b4d2e24a16cd0d9b0#diff-b37151fda8ddf2f86daeb0d2f62f3225e64ce14c0b7f1548cb2ef38bf891e85fR43

marosset: This gets set in the powershell snippet above (line 43).
I'm not sure what it does but hopefully someone else who can comment on the PR

kubeadm/scripts/Install-Containerd.ps1
```diff
@@ -0,0 +1,167 @@
...
+    "Containers",
+    "Hyper-V",
+    "Hyper-V-PowerShell")
+
+function ValidateWindowsFeatures {
+    $allFeaturesInstalled = $true
+    foreach ($feature in $requiredWindowsFeatures) {
+        $f = Get-WindowsFeature -Name $feature
+        if (-not $f.Installed) {
+            Write-Warning "Windows feature: '$feature' is not instealled."
```
jsturtevant, line 95:
```suggestion
            Write-Warning "Windows feature: '$feature' is not installed."
```

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -51,9 +69,17 @@ DownloadFile $kubeletBinPath https://dl.k8s.io/$KubernetesVersion/bin/windows/am
...
+    # Create host network to allow kubelet to schedule hostNetwork pods
+    # NOTE: For containerd the 0-containerd-nat.json network config template added by
+    # Install-containerd.ps1 joins pods to the host network.
+    Write-Host "Creating Docker host network"
+    docker network create -d nat host
+} elseif ($ContainerRuntime -eq "containerD") {
+    DownloadFile "c:\k\hns.psm1" https://github.com/Microsoft/SDN/raw/master/Kubernetes/windows/hns.psm1
+    Import-Module "c:\k\hns.psm1"
+    New-HnsNetwork -Type NAT -Name nat
```
jsturtevant, line 81: When running through this I got the following error on this step: 
```
Invoke-HnsRequest : @{Error=A network with this name already exists. ; ErrorCode=2151350288; Success=False}                                             At C:\k\hns.psm1:233 char:16                                                                                                                            + ...      return Invoke-HnsRequest -Method POST -Type networks -Data $Json .
```

I think this maybe because containerd started with the nat network config during install containerd?

jsturtevant: This did not cause a fatal error just a warning that it was already created.

marosset: I added a TODO in the code to come back and clean this up

kubeadm/scripts/Install-Containerd.ps1
marosset: fixed

kubeadm/flannel/flannel-overlay.yml
daschott: @jsturtevant yes this should be set to the ip of the host. 

I didn't intent for this comment to be blocking initially since I thought this would automatically be added as was the case previously. But upon attempting to remove this, pod-pod connectivity is failing, so this should be included.

Provider Address (PA) is the IP address assigned to the container hosts network infrastructure. The PA appears in the packets on the hosts network. Without the PA, there is no routing for Windows pods and network will not know how get traffic to the right destination host

jsturtevant:
I was able to join a node:

```
kubectl get nodes -o wide
NAME           STATUS   ROLES    AGE   VERSION   INTERNAL-IP   EXTERNAL-IP   OS-IMAGE                         KERNEL-VERSION     CONTAINER-RUNTIME
controller-0   Ready    master   19h   v1.19.3   10.240.0.10   <none>        Ubuntu 18.04.4 LTS               5.3.0-1028-azure   docker://19.3.13
worker-win-2   Ready    <none>   17h   v1.19.3   10.240.0.22   <none>        Windows Server 2019 Datacenter   10.0.17763.1397    containerd://1.4.1
```

PR #120, opened 2020-11-06
Fix StartKubelet.ps1 generation

vitaliy-leschenko:
Current variable $StartKubeletFileContent has incorrect string because it contains '.
We need to replace ' to '' to fix the issue.


marosset:
I'm going to file an issue to see if we can get some powershell linting/validation as CI checks for this repo.

marosset:
https://github.com/kubernetes-sigs/sig-windows-tools/issues/121

PR #125, opened 2020-11-19
Add note about tricky rancher file rename for kube-proxy.exe

jayunit100:
just a quick note for this kube-proxy rename thats been puzzling me for an hour or so :).  
Seems like processes are renamed on the host to `rancher-wins-*`.  
It seems like we may want to plumb this in as a variable to rancher-wins someday.  

Created https://github.com/rancher/wins/issues/11 , in case theres any value in **not** renaming the file on startup.


jayunit100:
added a note about how to live-patch kube-proxy

PR #131, opened 2020-12-02
DebugWindowsNode.ps1: fix docker version checking

vitaliy-leschenko:
**Reason for PR**:
Script DebugWindowsNode.ps1 has incorrect docker version checking.
Powershell `-contains` doesn't work with masks.
Previously script was failed with docker 19.03.13.

Now it is OK.

### DIFF
--- a/hack/DebugWindowsNode.ps1
+++ b/hack/DebugWindowsNode.ps1
@@ -57,13 +57,12 @@ Describe "Docker is installed" {
     }
 
     It "Should be a supported version" {
-        $supportedVersions = @(
-            "19.03*",
-            "19.04*",
-            "19.1*",
-            "20*"
-        )
-        $supportedVersions -contains $global:DockerVersion.Server.Version | Should Be $true
+        $supported = $false
+        if ($global:DockerVersion.Server.Version -match "\d+\.\d+\.\d+") {
+             $dockerVersion = [version]$Matches[0]
+             $supported = $dockerVersion -ge [version]"19.3.0"
+        }
+        $supported | Should Be $true
     }
 }
 


hack/DebugWindowsNode.ps1
```diff
@@ -57,13 +57,12 @@ Describe "Docker is installed" {
...
     It "Should be a supported version" {
-        $supportedVersions = @(
-            "19.03*",
-            "19.04*",
-            "19.1*",
-            "20*"
-        )
-        $supportedVersions -contains $global:DockerVersion.Server.Version | Should Be $true
+        $supported = $false
+        if ($global:DockerVersion.Server.Version -match "\d+\.\d+\.\d+") {
```
jsturtevant, line 61: is this additional matching required?  I think this whole if statement could be simplified to

```
$supported = [version]$global:DockerVersion.Server.Version -ge [version]"19.3.0"
```

if the version isn't semantic it will error

vitaliy-leschenko: Yes, it does.
```
PS C:\> [version]"20.10.0-rc1"
InvalidArgument: Cannot convert value "20.10.0-rc1" to type "System.Version". Error: "Input string was not in a correct format."
```

jayunit100:
ahh thanks !  i think ill just trusy you this since you ran into the bug and your the first person (other then me) to start using this script.
Please keep up the contributions ! hoping to add more fine grained info over time (lots of different corner cases we'll encounter w/ containerd, docker, and so on over time, so need to keep adding logic) 

Id say, lets merge and iterate

jayunit100 (approved):

hack/DebugWindowsNode.ps1
jsturtevant: interesting, looks like `[version]` doesn't handle the full sem ver spec

Issue #130, opened 2020-12-02
 PrepareNode.ps1 is not fit to run several times


Celthi:
**Is your feature request related to a problem? Please describe.**
I was following the page https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/adding-windows-nodes/ to run the `PrepareNode.ps1`.  But I failed the first time. After I fix the problem, I tried to run the script again to prepare the node. But now the script won't be able to run due to some settings like the docker network has been created.

**Describe the solution you'd like**
The script should be able to run multiple times and if a step is not needed due to finish in the previous run then it will just skip the step. So newbie can run the script to prepare the node after fix some errors.
**Describe alternatives you've considered**
Comment out some steps in the prepareNode.ps1 and run it again.
**Additional context**

PR #132, opened 2020-12-02
Allow to run PrepareNode.ps1 several times

vitaliy-leschenko:
**Issue Fixed**:
Issue #130 

**Notes**:
Added new script CleanupNode.ps1 to cleanup node

### DIFF
--- a/kubeadm/scripts/CleanupNode.ps1
+++ b/kubeadm/scripts/CleanupNode.ps1
@@ -12,13 +12,17 @@ if (Get-Service docker -ErrorAction Ignore) {
 if (Get-Service containerd -ErrorAction Ignore) {
     Restart-Service containerd
 }
-del /etc -Recurse -Force
-del /opt -Recurse -Force
-del /run -Recurse -Force
-del /var -Recurse -Force
-del /k/flannel -Recurse -Force
-del /k/kube-proxy -Recurse -Force
-del /k/*.exe
-del /k/StartKubelet.ps1
+Stop-Process -Name rancher-wins* -Force
 Get-HnsNetwork | ?{ $_.type -eq "L2Bridge" } | Remove-HnsNetwork
 Get-HnsNetwork | ?{ $_.type -eq "Overlay" } | Remove-HnsNetwork
+
+del /etc -Recurse -Force | out-null
+del /opt -Recurse -Force | out-null
+del /run -Recurse -Force | out-null
+cmd /c rmdir c:\var /S /Q | out-null
+del /k/flannel -Recurse -Force | out-null
+del /k/kube-proxy -Recurse -Force | out-null
+del /k/*.exe | out-null
+del /k/StartKubelet.ps1 | out-null
+
+Write-Host "Please restart computer to complete node cleanup."
--- a/kubeadm/scripts/CleanupNode.ps1
+++ b/kubeadm/scripts/CleanupNode.ps1
@@ -1,0 +1,24 @@
+if (Get-Service kubelet -ErrorAction Ignore) {
+    Stop-Service kubelet
+    sc.exe delete kubelet
+}
+if (Get-Service rancher-wins -ErrorAction Ignore) {
+    Stop-Service rancher-wins
+    sc.exe delete rancher-wins
+}
+if (Get-Service docker -ErrorAction Ignore) {
+    Restart-Service docker
+}
+if (Get-Service containerd -ErrorAction Ignore) {
+    Restart-Service containerd
+}
+del /etc -Recurse -Force
+del /opt -Recurse -Force
+del /run -Recurse -Force
+del /var -Recurse -Force
+del /k/flannel -Recurse -Force
+del /k/kube-proxy -Recurse -Force
+del /k/*.exe
+del /k/StartKubelet.ps1
+Get-HnsNetwork | ?{ $_.type -eq "L2Bridge" } | Remove-HnsNetwork
+Get-HnsNetwork | ?{ $_.type -eq "Overlay" } | Remove-HnsNetwork
--- a/kubeadm/scripts/PrepareNode.ps1
+++ b/kubeadm/scripts/PrepareNode.ps1
@@ -64,6 +64,7 @@ mkdir -force "$global:KubernetesPath"
 $env:Path += ";$global:KubernetesPath"
 [Environment]::SetEnvironmentVariable("Path", $env:Path, [System.EnvironmentVariableTarget]::Machine)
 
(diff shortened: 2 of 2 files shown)


kubeadm/scripts/PrepareNode.ps1
```diff
@@ -59,81 +59,97 @@ $global:PowershellArgs = "-ExecutionPolicy Bypass -NoProfile"
...
+        nssm set kubelet DependOnService docker
+    } elseif ($ContainerRuntime -eq "containerD") {
+        nssm set kubelet DependOnService containerd
+    }
+} else {
+    Write-Host "kubelet has been already installed"
 }
 
-New-NetFirewallRule -Name kubelet -DisplayName 'kubelet' -Enabled True -Direction Inbound -Protocol TCP -Action Allow -LocalPort 10250
+New-NetFirewallRule -Name kubelet -DisplayName 'kubelet' -Enabled True -Direction Inbound -Protocol TCP -Action Allow -LocalPort 10250 -ErrorAction Ignore
```
agowa338, line 155: How about writing it like this? This has the benefit of not catching unexpected errors.
```suggestion
if (-not [bool](Get-NetFirewallRule -Name "kubelet")) {
    New-NetFirewallRule -Name kubelet -DisplayName 'kubelet' -Enabled True -Direction Inbound -Protocol TCP -Action Allow -LocalPort 10250
}
```

kubeadm/scripts/CleanupNode.ps1
```diff
@@ -0,0 +1,28 @@
...
+    Stop-Service rancher-wins
+    sc.exe delete rancher-wins
+}
+if (Get-Service docker -ErrorAction Ignore) {
+    Restart-Service docker
+}
+if (Get-Service containerd -ErrorAction Ignore) {
+    Restart-Service containerd
+}
+Stop-Process -Name rancher-wins* -Force
```
agowa338, line 15: Isn't the executable just `wins.exe` and the process name therefore `wins`?
```suggestion
Stop-Process -Name wins -Force
```

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -59,81 +59,97 @@ $global:PowershellArgs = "-ExecutionPolicy Bypass -NoProfile"
...
-DownloadFile nssm.zip https://k8stestinfrabinaries.blob.core.windows.net/nssm-mirror/nssm-2.24.zip
-tar C $global:NssmInstallDirectory -xvf .\nssm.zip --strip-components 2 */$arch/*.exe
-Remove-Item -Force .\nssm.zip
+if (-not (Get-Service kubelet -ErrorAction Ignore)) {
+    DownloadFile "$global:KubernetesPath\kubelet.exe" https://dl.k8s.io/$KubernetesVersion/bin/windows/amd64/kubelet.exe
+
+    mkdir -force C:\var\log\kubelet
+    mkdir -force C:\var\lib\kubelet\etc\kubernetes
+    mkdir -force C:\etc\kubernetes\pki
+    New-Item -path C:\var\lib\kubelet\etc\kubernetes\pki -type SymbolicLink -value C:\etc\kubernetes\pki\
```
agowa338, line 116: Without `-Force` this fails if it already exists.
```suggestion
    New-Item -path C:\var\lib\kubelet\etc\kubernetes\pki -type SymbolicLink -value C:\etc\kubernetes\pki\ -Force
```

agowa338 (commented):
Wish I saw this PR earlier, but as I just now implemented the same functionality (in a lesser sophisticated way) I may as well just provide some feedback.

kubeadm/scripts/PrepareNode.ps1
vitaliy-leschenko: You are right. It is better.

kubeadm/scripts/CleanupNode.ps1
vitaliy-leschenko: Unfortunately no. In this case we try to cleanup this processes:
```
PS C:\Users\v.leschenko> Get-Process rancher-wins*

Handles  NPM(K)    PM(K)      WS(K)     CPU(s)     Id  SI ProcessName
-------  ------    -----      -----     ------     --  -- -----------
    228      15    21716      22048     484.08  16672   0 rancher-wins-flanneld
    231      19    29024      30968   1,583.41   8896   0 rancher-wins-kube-proxy
```

Sometimes when something happens this processed running even after rancher-wins service was removed.

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -59,81 +59,99 @@ $global:PowershellArgs = "-ExecutionPolicy Bypass -NoProfile"
...
-$StartKubeletFileContent = $StartKubeletFileContent -replace "{{CONTAINER_RUNTIME}}", "`"$ContainerRuntime`""
-Set-Content -Path $global:StartKubeletScript -Value $StartKubeletFileContent
+    mkdir -Force $global:NssmInstallDirectory | Out-Null
+    DownloadFile nssm.zip https://k8stestinfrabinaries.blob.core.windows.net/nssm-mirror/nssm-2.24.zip
+    tar C $global:NssmInstallDirectory -xvf .\nssm.zip --strip-components 2 */$arch/*.exe
+    Remove-Item -Force .\nssm.zip
+
+    $env:path += ";$global:NssmInstallDirectory"
+    $newPath = "$global:NssmInstallDirectory;" +
+    [Environment]::GetEnvironmentVariable("PATH", [EnvironmentVariableTarget]::Machine)
```
agowa338, line 98: This will prefix system path with the nssm install directory. For now this shouldn't be an issue. But it is bad practice, as it will slow down dynamic dll imports as this directory will get checked first every time. Also it will allow system executables and libraries to be accidentally overwritten.

```suggestion
    $newPath = [Environment]::GetEnvironmentVariable("PATH", [EnvironmentVariableTarget]::Machine) + ";$global:NssmInstallDirectory"
```

vitaliy-leschenko: It makes sense.

jsturtevant:
Thanks looking at this.  @agowa338 Could you give a review since you've run into this as well?  It seems like your PR had fewer changes.

@vitaliy-leschenko I have a concern about having this as 2 files.  The docs have you download 2 files.  Does cleanup node need to be run if this is run twice?

vitaliy-leschenko:
> I have a concern about having this as 2 files. The docs have you download 2 files. Does cleanup node need to be run if this is run twice?

No, it doesn't. We can run PrepareNode.ps1 many times without run cleanup. But if we want to leave the cluster and remove all artifacts we need to use CleanupNode.ps1. Also it maybe useful for some troubleshooting.

Issue #138, opened 2021-01-20
kubeadm job on GCE is failing


neolit123:
**Describe the bug**
the kubeadm job on GCE is failing
https://k8s-testgrid.appspot.com/sig-windows-gce#kubeadm-windows-gcp-k8s-stable

**To Reproduce**
one should have GCE service account and run the scripts locally from the /kubeadm folder in this repo.

**Expected behavior**
the test not to fail.

**Kubernetes (please complete the following information):**
 - Windows Server version: not sure
 - Kubernetes Version: latest stable
 - CNI: flannel

**Additional context**
can try looking at why this week (or more likely the next one).

/sig windows cluster-lifecycle
/kind failing-test
/assign

neolit123:
the job deployment is now working:
https://k8s-testgrid.appspot.com/sig-windows-releases#kubeadm-windows-gcp-k8s-stable

some conformance test appear to be red, which is not deployer related.

/close

PR #140, opened 2021-01-25
kubeadm: write "cgroupfs" in the kubelet config

neolit123:
**Reason for PR**:

Cgroup drivers should be ignored on Windows but there is a bug
in the kubelet:
  kubernetes/kubernetes#97759
the bug is fixed in master (.21) but is not part of the latest stable releases in the skew yet.

Workaround this bug by patching the kubelet configuration in
the cluster.

**Issue Fixed**:

fixes https://github.com/kubernetes-sigs/sig-windows-tools/issues/138

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests

**Notes**:

### DIFF
--- a/kubeadm/hack/e2e-cluster-gcp.sh
+++ b/kubeadm/hack/e2e-cluster-gcp.sh
@@ -92,6 +92,18 @@ sed -i "s/${privateIP}/${externalIP}/g" $KUBECONFIG
 k8sVersion=$(kubectl version -ojson | jq -r .serverVersion.gitVersion)
 sed "s/VERSION/${k8sVersion}/" "$REPO_ROOT/kubeadm/hack/startup/windows.ps1" > $scratchDir/windows.ps1
 
+# Patch the cgroup driver in the kubelet config to be cgroupfs.
+# Workaround for https://github.com/kubernetes/kubernetes/issues/97759
+# This makes the joining Windows workers download a kubelet configuration that is compatible
+# with the Docker defaults on Windows.
+k8sVersionMajor=$(echo $k8sVersion | cut -d "." -f 1 | cut -c2-)
+k8sVersionMinor=$(echo $k8sVersion | cut -d "." -f 2)
+kubeletConfigName="kubelet-config-$k8sVersionMajor.$k8sVersionMinor"
+kubeletTmpFile="/tmp/kubelet.yaml"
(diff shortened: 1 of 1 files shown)


neolit123:
> the bug is fixed in master (.21) but is not part of the latest stable releases in the skew yet.

we should have backports for this:
https://github.com/kubernetes/kubernetes/pull/97764#issuecomment-767016913

jsturtevant:
I am not super familiar with this setup.  It looks like the cherrypicks went in for older versions: https://github.com/kubernetes/kubernetes/pull/98384

Should we prefer those over the patch?

neolit123:
@jsturtevant 

i've sent the PR after investigating the failures and so that we don't delete the job due to not being maintained.
cc @jayunit100 

these scripts were originally created by benmoss with some help from me, but ideally sig-windows should take ownership.
your observation about the cherry picks is correct - once a new k8s PATCH release is out for the latest stable (1.20) this setup should no longer fail and this patch will not be needed.

we missed v1.20.2 for the cherry pick, so it will be out with v1.20.3, maybe in ~2 weeks.
/hold

jsturtevant:
Sounds good. FWIW we have additional coverage for windows Kubeadm over in capz: https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-azure#periodic-capi-e2e-full  

It's not as extensive (I am[ working on that](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1119)) but it does validate that kubeadm binary works. I'll need to make sure we get the correct fix there too.

neolit123:
given capz is more actively maintained that the CI scripts in this repository, we can remove them and if that is what sig-windows would prefer.

neolit123:
1.20.4 was just released:
https://groups.google.com/d/msgid/kubernetes-dev/pYK9msu4TUqsFh7mUQiCXQ%40geopod-ismtpd-1-0?utm_medium=email&utm_source=footer

will check if the latest kubelet there fixes the related failures here:
https://k8s-testgrid.appspot.com/sig-windows-gce#kubeadm-windows-gcp-k8s-stable

neolit123:
the job deployment is now working:
https://k8s-testgrid.appspot.com/sig-windows-releases#kubeadm-windows-gcp-k8s-stable

some conformance test appear to be red, which is not deployer related.

/close

PR #144, opened 2021-03-02
Make PrepareNode idempotent

### LINKED ISSUE #130: PrepareNode.ps1 is not fit to run several times

agowa338:
Making PrepareNode idempotent, so that deployment scripts that may want to run this script at every startup have an easier job.

**Reason for PR**:
- There was todo for checking before creating within the code. That's fixed now.
- It makes my deployment script simpler if I don't need to deal with the missing idempotency of this script at runtime.
- I didn't notice that (#132) already existed.

**Issue Fixed**:

Fixes #130
Fixes #132

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests

**Notes**:

### DIFF
--- a/kubeadm/scripts/PrepareNode.ps1
+++ b/kubeadm/scripts/PrepareNode.ps1
@@ -65,6 +65,9 @@ mkdir -force "$global:KubernetesPath"
 $env:Path += ";$global:KubernetesPath"
 [Environment]::SetEnvironmentVariable("Path", $env:Path, [System.EnvironmentVariableTarget]::Machine)
 
+if ([bool](Get-Service -Name "rancher-wins")) { Stop-Service -Name "rancher-wins" }
+if ([bool](Get-Service -Name "kubelet")) { Stop-Service -Name "kubelet" }
+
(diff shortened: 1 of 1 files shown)


vitaliy-leschenko:
What do you think about combine our PRs (#144, #132) into one?

agowa338:
I'm going to close this one as you already integrated my changes via the suggestions I made. Therefore this shouldn't be needed anymore.
If I've overlooked something, feel free to just c'n'p it over.

PR #147, opened 2021-04-12
Explicity disable dualstack for win overlay

jsturtevant:
**Reason for PR**:

1.21 broke win overlay with dualstack: https://github.com/kubernetes/kubernetes/issues/100966

The kubeadm tests are broken becuase of this:
https://testgrid.k8s.io/sig-windows-master-release#kubeadm-windows-gcp-k8s-stable

**Issue Fixed**:

Issue #

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests

**Notes**:

/assign @marosset @neolit123

### DIFF
--- a/kubeadm/kube-proxy/kube-proxy.yml
+++ b/kubeadm/kube-proxy/kube-proxy.yml
@@ -43,6 +43,7 @@ data:
     yq w -i /host/var/lib/kube-proxy/config.conf winkernel.sourceVip $sourceVip
     yq w -i /host/var/lib/kube-proxy/config.conf winkernel.networkName $networkName
     yq w -i /host/var/lib/kube-proxy/config.conf featureGates.WinOverlay true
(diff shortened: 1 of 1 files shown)


neolit123:
given this is a breaking change in kube-proxy we should be fixing it in a 1.21 patch release?
to me it seems we should wait for that, instead of toggling off the feature gate.

jsturtevant:
agree they should be fixed in a patch release, but win overlay isn't supported with dualstack either way.  While waiting for the patch release I would rather have the tests running to catch other issues that might be introduced

neolit123:
ok, understood.
/lgtm
/hold

cancel the hold if you think this should merge now.

jsturtevant:
PR to for fixing the patch: https://github.com/kubernetes/kubernetes/pull/101047
Issue to track rolling this back:  #148
/hold cancel

PR #151, opened 2021-06-07
Update PrepareNode.ps1

jayunit100:
Quick fix to enable downloads


neolit123:
Probably better to download if the local file has a different checksum, to
avoid version mismatch.

PR #161, opened 2021-08-19
Add Host process examples

jsturtevant:
**Reason for PR**:

Start providing some examples of how to use hostprocess

**Issue Fixed**:

Issue #

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests

**Notes**:

This is meant to be an example to start testing things out.  Not ready for production.


hostprocess/flannel/build.sh
```diff
@@ -0,0 +1,3 @@
+docker buildx create --name img-builder --use --platform windows/amd64
+trap 'docker buildx rm img-builder' EXIT
+docker buildx build --platform windows/amd64 --output=type=registry --pull -f Dockerfile -t jsturtevant/flannel:hostprocess .
```
marosset, line 3: Let's not hardcode your dockerhub repo :)

hostprocess/flannel/flannel-overlay.yml
```diff
@@ -0,0 +1,136 @@
...
+                    values:
+                      - windows
+                  - key: kubernetes.io/arch
+                    operator: In
+                    values:
+                      - amd64
+      securityContext:
+        windowsOptions:
+          hostProcess: true
+          runAsUserName: "NT AUTHORITY\\system"
```
marosset, line 89: Can we use `NT AUTHORITY\\NetworkService` here?

hostprocess/flannel/flannel-overlay.yml
```diff
@@ -0,0 +1,136 @@
...
+      - operator: Exists
+        effect: NoSchedule
+        # Mark the pod as a critical add-on for rescheduling.
+      - key: CriticalAddonsOnly
+        operator: Exists
+      - effect: NoExecute
+        operator: Exists
+      containers:
+      - name: kube-flannel
+        image: jsturtevant/flannel-tools:v0.14.0-1809-hostprocess
```
marosset, line 102: use sigwindowstools?

hostprocess/flannel/build.sh
jsturtevant: agree, before I pushed to the sig-windows-tools repo I wanted to get general agreement in this direction.

hostprocess/flannel/flannel-overlay.yml
jsturtevant: It is on my list to try this.  `local system` did not work because calls to HNS apis failed silently

hostprocess/flannel/flannel-overlay.yml
jsturtevant: before I pushed to the sig-windows-tools repo I wanted to get general agreement in this direction.

hostprocess/flannel/build.sh
jsturtevant: fixed

hostprocess/calico/calico.yml
```diff
@@ -0,0 +1,238 @@
...
+      - key: CriticalAddonsOnly
+        operator: Exists
+      - effect: NoExecute
+        operator: Exists
+      initContainers:
+        # This container installs the CNI binaries
+        # and CNI network config file on each node.
+        - name: install-cni
+          image: sigwindowstools/calico-install:v3.20.0-hostprocess
+          args: ["$env:CONTAINER_SANDBOX_MOUNT_POINT/calico/install.ps1"]
```
liggitt, line 135: what interprets this envvar? normally, envvars would be specified in args as `$(CONTAINER_SANDBOX_MOUNT_POINT)/calico/install.ps1` (https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config)

hostprocess/calico/calico.yml
```diff
@@ -0,0 +1,238 @@
...
+              mountPath: /etc/kubeadm-config/
+          securityContext:
+            windowsOptions:
+              hostProcess: true
+              runAsUserName: "NT AUTHORITY\\system"
+      containers:
+      - name: calico-node-startup
+        image: sigwindowstools/calico-node:v3.20.0-hostprocess
+        args: ["$env:CONTAINER_SANDBOX_MOUNT_POINT/calico/node-service.ps1"]
+        workingDir: "$env:CONTAINER_SANDBOX_MOUNT_POINT/calico/"
```
liggitt, line 178: same question about what component evaluates/expands this envvar... are envvars normally usable inside workingDir?

it looks like the kubelet only expands envvars in command/args, and workingdir is passed as-is to the runtime:

https://github.com/kubernetes/kubernetes/blob/95390e6476fca15b0f55cde2c142ec18ca57aee0/pkg/kubelet/kuberuntime/kuberuntime_container.go#L313-L329

marosset: These variables are set for processes created inside the container by hcsshim (for both normal and hostProcess containers). Windows will expand and environment variables found in the command line when it tries to runt he process / entry point for the container.

jsturtevant: The runtime sets and evaluates these ENV variables as part of its path processing. We can't use the approach in https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config)  since it is not known until runtime since that value is set by hcsschim.

Using the vars, technically shouldn't be necessary for hostprocess but we found some bugs  in hcsshim around arg and Working directory path processing.  With the approach that we are evaluating now this env variable won't be necessary any longer as the container will be given its own filesystem (thought it still has access to the host filesystem unlike on linux).

For reference these are the fixes that are in being worked on until the new approach is validated:
https://github.com/microsoft/hcsshim/pull/1137
https://github.com/microsoft/hcsshim/pull/1117

cc: @dcantah

hostprocess/flannel/flannel-overlay.yml
perithompson: The network service doesn't work either, I think there were some folder permissions or something, I can't quite remember but I was hoping to use network service as well!

jsturtevant: I'll give this a try and see if I get an error message. @dcantah @kevpar any thoughts on what specific permissions are needed?

dcantah: No not off the top of my head, I can ask some hns folks

hostprocess/build.sh
```diff
@@ -0,0 +1,26 @@
...
+ 
+# Read the array values with space
+for proxyVersion in "${proxyVersions[@]}"; do
+  pushd $SCRIPTROOT/flannel
+  ./build.sh -r $repository --proxyVersion $proxyVersion
+  popd
+  pushd $SCRIPTROOT/calico
+  ./build.sh -r $repository --proxyVersion $proxyVersion
+  popd
+done
```
marosset, line 26:
```suggestion
done

```

hostprocess/calico/install/Dockerfile.install
```diff
@@ -0,0 +1,22 @@
...
+FROM $BASE
+
+ENV PATH="C:\Program Files\PowerShell;C:\utils;C:\Windows\system32;C:\Windows;"
+
+COPY --from=bins /cni/CalicoWindows/cni/calico.exe /calico/cni/calico.exe
+COPY --from=bins /cni/CalicoWindows/cni/calico-ipam.exe /calico/cni/calico-ipam.exe
+COPY --from=bins /cni/CalicoWindows/calico-kube-config.template /calico/calico-kube-config.template
+
+COPY calico-install.ps1 /calico/install.ps1
+ENTRYPOINT ["powershell"]
```
marosset, line 22:
```suggestion
ENTRYPOINT ["powershell"]

```

hostprocess/calico/install/calico-install.ps1
```diff
@@ -0,0 +1,35 @@
...
+Copy-Item -Path "$env:CONTAINER_SANDBOX_MOUNT_POINT\calico\cni\calico-ipam.exe" -Destination "$env:CNI_BIN_DIR\calico-ipam.exe" -Force
+
+Write-Host "Writing CNI configuration to $env:CNI_CONF_DIR."
+(echo "$env:CNI_NETWORK_CONFIG").
+    replace('__KUBECONFIG_FILEPATH__', "$env:CNI_CONF_DIR/calico-kubeconfig").
+    replace('__K8S_SERVICE_CIDR__', $env:K8S_SERVICE_CIDR).
+    replace('__KUBERNETES_NODE_NAME__', $env:KUBERNETES_NODE_NAME).
+    replace('__CNI_MTU__', $env:CNI_MTU) | Set-Content "${env:CNI_CONF_DIR}/10-calico.conflist" -Force
+
+Write-Host "CNI configured"
```
marosset, line 35:
```suggestion
Write-Host "CNI configured"

```

marosset:
Can we update the 1.22.1 references to 1.22.2?

hostprocess/flannel/flanneld/start.ps1
```diff
@@ -0,0 +1,42 @@
...
+write-host "copy sa info (should be able to do this with a change to go client"
+mkdir -force $env:CONTAINER_SANDBOX_MOUNT_POINT/flannel-config-file/var/run/secrets/kubernetes.io/serviceaccount/
+cp -force $env:CONTAINER_SANDBOX_MOUNT_POINT/var/run/secrets/kubernetes.io/serviceaccount/* $env:CONTAINER_SANDBOX_MOUNT_POINT/flannel-config-file/var/run/secrets/kubernetes.io/serviceaccount/
+
+Write-Host "envs"
+write-host $env:POD_NAME
+write-host $env:POD_NAMESPACE
+
+Write-Host "Starting flannel"
+& $env:CONTAINER_SANDBOX_MOUNT_POINT/flannel/flanneld.exe --kube-subnet-mgr --kubeconfig-file $env:CONTAINER_SANDBOX_MOUNT_POINT/flannel-config-file/kubeconfig.conf --iface 10.1.0.5
```
marosset, line 42:
```suggestion
& $env:CONTAINER_SANDBOX_MOUNT_POINT/flannel/flanneld.exe --kube-subnet-mgr --kubeconfig-file $env:CONTAINER_SANDBOX_MOUNT_POINT/flannel-config-file/kubeconfig.conf --iface 10.1.0.5

```

hostprocess/calico/node/felix-service.ps1
```diff
@@ -0,0 +1,39 @@
...
+
+# Autoconfigure the IPAM block mode.
+if ($env:CNI_IPAM_TYPE -EQ "host-local") {
+    $env:USE_POD_CIDR = "true"
+} else {
+    $env:USE_POD_CIDR = "false"
+}
+
+# Run the calico-felix binary.
+.\calico-node.exe -felix
```
marosset, line 39:
```suggestion
.\calico-node.exe -felix

```

marosset:
> Can we update the 1.22.1 references to 1.22.2?

Other than that I think this is ready to merge (and we can iterate later if needed)

PR #163, opened 2021-08-23
No longer need HostProcess specific build since containerd pr merged

jsturtevant:
**Reason for PR**:

Removes the HostProcess build as https://github.com/containerd/containerd/pull/5131 has merged.

**Issue Fixed**:

Issue #

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests

**Notes**:


marosset:
Can we just disable it for now?
I think we're going to want to be able to target dev branches for some work in the future.

jsturtevant:
I lean toward removing it to avoid confusion several months later when we revisit if the requirement for dev branches is required.  This was really to fill a gap while we landed it across all of the moving parts. The less custom stuff we have the better IMO.

marosset:
Can we keep this around (but disabled) until the KEP to move HostProcess containers to beta gets reviewed (Sept 9 is the enhancement freeze)?
After the KEP review we'll have a much better idea what kinds of changes we'll be making for hcsshim and/or containerd and if those changes would justify keeping the forked builds around?

marosset (approved):
/lgtm

I spoke w/ @jsturtevant offline and we decided we'll just bring  back the workflow if needed to keep the repo tidy.

PR #183, opened 2021-11-04
Use pwsh if available

craigminihan:
**Reason for PR**:
On `mcr.microsoft.com/powershell:lts-nanoserver-20h2` the PowerShell exe is `pwsh.exe` not `powershell.exe`. This image is much lighter than the alternate `windowsservercore` and is preferable for deployment to nodes running Windows 20h2.

**Notes**:
* Change is non-breaking since it will fallback to `powershell.exe`.
* nanoserver requires `netapi32.dll` to be copied from a matching `windowsservercore` image
* original source file has a mix of spaces and tabs

### DIFF
--- a/kubeadm/flannel/setup.go
+++ b/kubeadm/flannel/setup.go
@@ -33,7 +33,13 @@ func setupL2bridge(interfaceName string) {
 }
 
 func run(command string) {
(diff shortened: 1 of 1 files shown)


craigminihan:
I signed it

PR #188, opened 2021-11-28
fix(hostprocess-calico): Prevent removing WSL network

lippertmarkus:
**Reason for PR**:
This prevents the calico node startup script to delete the WSL HnsNetwork which breaks WSL when running on Windows 11.

**Issue Fixed**:

**Requirements**

- [x] Sqaush commits 
- [-] Documentation
- [-] Tests

**Notes**:

### DIFF
--- a/hostprocess/calico/node/node-service.ps1
+++ b/hostprocess/calico/node/node-service.ps1
@@ -41,15 +41,15 @@ if ($env:CALICO_NETWORKING_BACKEND -EQ "windows-bgp" -OR $env:CALICO_NETWORKING_
     $prevLastBootTime = Get-StoredLastBootTime
     if ($prevLastBootTime -NE $lastBootTime)
     {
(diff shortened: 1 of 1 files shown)


jsturtevant:
Are you running this on Windows 10/11?

lippertmarkus:
Yes, tried out a setup where the control plane is running in WSL and the Windows 11 host joins as a Worker. Thought that might be a good development setup somewhere in the future. Existing solutions like Rancher Desktop don't support adding Windows nodes and especially when you develop apps for Windows Containers you often need some other components running on Linux.

jsturtevant:
> Yes, tried out a setup where the control plane is running in WSL and the Windows 11 host joins as a Worker. Thought that might be a good development setup somewhere in the future.

this is a great idea! have you gotten it to work?

I don't see any harm in updating these manifests as it is a no op if it isn't there and opens up the door for this scenario.

marosset (approved):
/lgtm

PR #190, opened 2021-11-30
use latest ubuntu to avoid CVEs

jayunit100:
quick patch to use ubuntu, to avoid CVEs/ old alpine versions

### DIFF
--- a/burrito/Dockerfile
+++ b/burrito/Dockerfile
@@ -1,5 +1,5 @@
 ARG BUILDER_IMAGE=golang:1.16-alpine
-ARG RUN_IMAGE=alpine:latest
+ARG RUN_IMAGE=ubuntu:latest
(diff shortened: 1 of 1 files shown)


rikatz:
/lgtm
Thanks

burrito/Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG BUILDER_IMAGE=golang:1.16-alpine
```
PushkarJ, line 1: Should we use golang:1.16 without alpine? i.e. this image: https://hub.docker.com/layers/golang/library/golang/1.16/images/sha256-f1cec27767a33f48aaf376d69815f52c9b8202c2d1fd24a7a7778b68f43d1b2f?context=explore for consistency

PR #195, opened 2021-12-16
Add nightly build for hostprocess images

vitaliy-leschenko:
**Reason for PR**:
Add nightly build for hostprocess images:
- sigwindowstools/flannel:{flannelVersion}-hostprocess
- sigwindowstools/calico-install:{calicoVersion}-hostprocess
- sigwindowstools/calico-node:{calicoVersion}-hostprocess
- sigwindowstools/kube-proxy:{k8sVersion}-flannel-hostprocess
- sigwindowstools/kube-proxy:{k8sVersion}-calico-hostprocess

where:
- k8sVersion - all k8s versions from 1.22.0 (as min version that supports hostprocess)
- calicoVersion - all calico version from 3.19.0 (as min version that supports windows)
- flannelVersion - all flannel version from 0.12.0

### DIFF
--- a/hostprocess/build-calico.ps1
+++ b/hostprocess/build-calico.ps1
@@ -1,0 +1,46 @@
+param(
+    [string]$repository = "sigwindowstools",
+    [version]$minCalicoVersion = "3.19.0",
+    [version]$minK8sVersion = "1.22.0"
+)
+
+pushd calico
+write-host "build calico"
+$calicoVersions = (curl -L https://api.github.com/repos/projectcalico/calico/releases | ConvertFrom-Json) | % tag_name
+foreach($calicoVersion in $calicoVersions)
+{
+    if ($calicoVersion -match "^v(\d+\.\d+\.\d+)$")
+    {
+        $testVersion = [version]$Matches[1]
+        if ($testVersion -ge $minCalicoVersion)
+        {
+            Write-Host "Build images for calico $calicoVersion"
+            pushd install
+            docker buildx build --platform windows/amd64 --output=type=registry --pull --build-arg=CALICO_VERSION=$calicoVersion -f Dockerfile.install -t $repository/calico-install:$calicoVersion-hostprocess .
+            popd
+            pushd node
+            docker buildx build --platform windows/amd64 --output=type=registry --pull --build-arg=CALICO_VERSION=$calicoVersion -f Dockerfile.node -t $repository/calico-node:$calicoVersion-hostprocess .
+            popd
+        }
+    }
+}
+
+
+write-host "build kube-proxy"
+pushd kube-proxy
+$versions = (curl -L k8s.gcr.io/v2/kube-proxy/tags/list | ConvertFrom-Json).tags
+foreach($version in $versions)
+{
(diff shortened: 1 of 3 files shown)


hostprocess/build-calico.ps1
```diff
@@ -0,0 +1,46 @@
+param(
+    [string]$repository = "sigwindowstools",
+    [version]$minCalicoVersion = "3.19.0",
+    [version]$minK8sVersion = "1.22.0"
+)
+
+pushd calico
+write-host "build calico"
+$calicoVersions = (curl -L https://api.github.com/repos/projectcalico/calico/releases | ConvertFrom-Json) | % tag_name
+foreach($calicoVersion in $calicoVersions)
```
jsturtevant, line 10: I have not tested with any of the other calico versions... I did modify the startup scripts to get this to work with hostprocess.  I am hesitant to add more calico versions without further investigation.

jsturtevant (commented):
Thanks for doing this, it has been on my list but i've been a bit hesitant to since these are mostly samples for inspiration.

Any reason to not use https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/hostprocess/build.sh?  This is the file I use to push the samples.  For 1.23, I added a new entry here: https://github.com/kubernetes-sigs/sig-windows-tools/blob/42d4411003b94e086356f891b278d452fc8f50e8/hostprocess/build.sh#L16 which i haven't pushed yet since 1.23.1 is coming out today and was waiting to add that.

hostprocess/build-calico.ps1
vitaliy-leschenko: I used `pwsh` just because have no experience to write bash scripts and don't know how to code the same script but using bash.

vitaliy-leschenko: I understand that we can't test all pairs `calico` and `kube-proxy` but anyway these may contain some issues and only user can decide which versions better for him. So, my proposal to allow all of them.

As additional point it makes possible to upgrade k8s cluster to any public version include minor versions.

vitaliy-leschenko: I have tested the last of calico on the last of kubernetes. For me it works very good.

jsturtevant: Thanks for doing this and testing the various versions.  

I thought about this some more and at this time I don't want to add all the different versions matrixes. As you point these are more demonstrations of how this works and users should be able to adjust them to their own needs pretty easily.  I want to avoid getting into debugging CNI configurations in this repo as we did the files in the kubeadm folder.  We don't have the breadth of CNI knowledge to maintain multiple cni's and versions.

The long-term goal for kube-proxy is move it k/k repo as an official image.  I will be working on that in the coming weeks.

For the CNI's, these should be hosted by cni providers longer term. I started to upstream the work for Calico but due to some limitations in the host-process design as of now it wasn't a good solution.  We are re-working the underlying solution so it will be easier to integrate and that should be coming in the next couple months.

That said we can maintain a few limited options in the meantime and I will try to add a job using the scripts we have currently.  Thanks again for your continued contributions and feedback. I hope you keep it coming.

PR #203, opened 2022-01-12
Csi proxy image

marosset:
**Reason for PR**:

This PR builds/releases a csi-proxy container image.
This image is hosted here because running csi-proxy is a short-term solution (the longterm solution is to have the plugins consume csi-proxy as a library) and csi-proxy repo did not want to maintain an official image.

**Issue Fixed**:

Issue #

**Requirements**

- [x] Sqaush commits 
- [x] Documentation
- [ ] Tests

**Notes**:


marosset:
This work started in https://github.com/kubernetes-csi/csi-proxy/pull/179

.github/workflows/csi-proxy.yml
```diff
@@ -0,0 +1,42 @@
+name: csi-proxy
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - hostprocess/csi-proxy/**
```
jsturtevant, line 7: This should be version file only otherwise it will trigger a push on every file change committed

hostprocess/csi-proxy/build.sh
```diff
@@ -0,0 +1,28 @@
...
+    exit 1
+fi
+echo "Using version ${version}"
+
+repository=${repository:-"ghcr.io/kubernetes-sigs/sig-windows"}
+
+docker buildx create --name img-builder --use --platform windows/amd64
+trap 'docker buildx rm img-builder' EXIT
+
+docker buildx build --platform windows/amd64 --output=type=registry -f Dockerfile.windows --build-arg=WINDOWS_VERSION=1809 -t ${repository}/csi-proxy:${version}-1809 .
```
jsturtevant, line 16: looks like we should be adding `--build-arg=CSI_PROXY_VERSION=$version` here to make sure we clone the correct csi proxy version

marosset: I thought about this some... Should we have anything CSI_PROXY_VERSION file in the repo for this?
I can see a need for having the image version and csi-proxy version not stay in sync (adding new windows OS versions for example).

jsturtevant: We need a way to bump the csi version.  The way this is done in k/k test images is use the same version but append a `-1` to each change.  so for instance if we made a change the image but still using 1.0.1 version of kube-proxy the image tag would be `1.0.1-2` where `2` is the version we are building.  This isn't perfect but for our purposes should work?

marosset: Right now I do have the csi-proxy version and the container image version decoupled.
The VERSION file is used for the container image version (ex `v1.0.2-2`) and the version of csi-proxy used is specified in the dockerfile.
Is this sufficient or do you think we should not have the version of csi-proxy used hardcoded in the Dockerfile?

.github/workflows/csi-proxy.yml
marosset: updated

hostprocess/csi-proxy/build.sh
jsturtevant: keeping it hard coded is probably fine we shouldn't support to many versions of the matrix

hostprocess/csi-proxy/Dockerfile.windows
```diff
@@ -0,0 +1,15 @@
...
+FROM --platform=linux/amd64 golang:1.16 as builder
+ARG CSI_PROXY_VERSION=v1.0.2
+RUN git clone https://github.com/kubernetes-csi/csi-proxy.git /go/csi-proxy &&\
+    cd /go/csi-proxy &&\
+    git checkout tags/${CSI_PROXY_VERSION} &&\
+    make build
+
+FROM ${REGISTRY}/${WINDOWS_BASE_IMAGE}:${WINDOWS_VERSION}
+COPY --from=builder /go/csi-proxy/bin/csi-proxy.exe /csi-proxy.exe
+ENV PATH="C:\Windows\system32;C:\Windows;C:\WINDOWS\System32\WindowsPowerShell\v1.0\;"
```
zhiweiv, line 14: Seems the env issue mentioned in https://github.com/kubernetes/kubernetes/blob/master/build/pause/Dockerfile_windows#L26-L28 has been fixed, is this sill needed?

jsturtevant: it might not be.

.github/workflows/csi-proxy.yml
```diff
@@ -0,0 +1,42 @@
...
+          echo "version=ci" >> $GITHUB_ENV
+      - name: set 'version' for release
+        if: ${{ github.event_name != 'pull_request' }}
+        run: |
+          export version=$(cat ./hostprocess/csi-proxy/VERSION)
+          echo "version=$version" >> $GITHUB_ENV
+      - name: build image
+        run: |
+          cd hostprocess/csi-proxy
+          ./build.sh --push
```
jsturtevant, line 42: Should there be one more step here? one for build and one for push. Where push only runs `if: ${{ github.event_name != 'pull_request' }}`

marosset: OOps, i think i meant to call `./build.sh` (without --push) for pull_requests and call `.build.sh --push) for non pull_requests.

marosset: Updated

PR #225, opened 2022-05-26
Remove feature gates in hostprocess kube-proxy script

Parkhost:
**Reason for PR**:

Removes feature gates in the kube-proxy hostprocess script: 

- winOverlay which is set to true since 1.20
- IPv6DualStack which is GA since 1.23


Parkhost:
not sure if setting; `IPv6DualStack` to `true` is breaking stuff.. maybe do a nightly build first..

jonaskello:
Would be great to have this merged. We've upgraded to k8s 1.24 but need to leave kube-proxy at v1.22 because of this.

jonaskello:
Thanks, upgraded and it works now! :-)

PR #236, opened 2022-08-09
Guide for Adding Windows Node

fabi200123:
**Reason for PR**:
This PR adds a guide for new users to help them add a Windows node to their kubernetes cluster.
This guide is based on the [Adding Windows nodes](https://kubernetes-docsy-staging.netlify.app/docs/tasks/administer-cluster/kubeadm/adding-windows-nodes/) page from the kubernetes documentation.

**Note**
It will be improved in the future to work for flannel hostprocess and calico hostprocess, but now it is only for flannel with rancher.

### DIFF
--- a/guides/guide-for-adding-windows-node.md
+++ b/guides/guide-for-adding-windows-node.md
@@ -1,0 +1,229 @@
+
+# Adding Windows nodes
+
+  
+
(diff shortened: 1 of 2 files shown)


guide-for-adding-windows-node.md
```diff
@@ -0,0 +1,229 @@
+
+# Adding Windows nodes
```
marosset, line 2: Can you create a new top-level folder named either `docs` or `guides` to put this in?

marosset: other than this comment LGTM.
This is a good starting point for docs thanks @fabi200123 !

PR #238, opened 2022-08-12
Solve an issue where serviceaccount is not created

fabi200123:
**Reason for PR**:
There is an issue that the flannel-overlay.yml does not create the **ServiceAccount flannel** and needed to be manually created and added it to the kube-flannel-rbac.yml to fix the issue. Also updated the readiness of the guide, there was an issue with a part of the code which identation was bad as well as a command that was supposed to get the pods for the namespace -kube-system (it should have been namespace -kube-flannel)

### DIFF
--- a/guides/guide-for-adding-windows-node.md
+++ b/guides/guide-for-adding-windows-node.md
@@ -1,60 +1,40 @@
 
 # Adding Windows nodes
 
-  
-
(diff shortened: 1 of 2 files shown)


claudiubelu:
This is mostly a cleanup, making the guide look a bit cleaner and adding the missing service account for flannel.

/lgtm
/approve

marosset (approved):

PR #212, opened 2022-04-11
Updates for kubeadm docs

jsturtevant:
**Reason for PR**:

Makes some updates to make the kubeadm docs easier.  

**Issue Fixed**:

Issue #

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests

**Notes**:

/sig windows

PR #239, opened 2022-08-12
Adding guide for calico and flannel hostprocess

fabi200123:
**Reason for PR**:
Added a section for flannel hostprocess and calico hostprocess. Flannel hostprocess needed as well as flannel with rancher, the same rbac file. For calico hostprocess a new rbac file needed to be created. 
The files from hostprocess/flannel were modified based on the PR: [#212 ](https://github.com/kubernetes-sigs/sig-windows-tools/pull/212) in order for them to work


guides/guide-for-adding-windows-node.md
```diff
@@ -1,229 +1,402 @@
...
-  
-
 Once you have a Linux-based Kubernetes control-plane node you are ready to choose a networking solution.
 
-  
 
-#### Configuring Flannel with rancher
-
-  
+## Configuring Flannel with rancher
```
jsturtevant, line 30: This should work on any base distro not just Rancher.  I don't see a reference to rancher beyond this.  To make sure folks who aren't using rancher don't turn away we should make this more generic:

```suggestion
## Configuring Flannel on Linux nodes
```
of something similiar?

jsturtevant (commented):
Thanks for adding this!  

@vrapolinario @lippertmarkus @jonaskello  you've all been through various guides and have good insights.  Would you take a look at these changes?

vrapolinario:
I'll review this - is there a timeline/ETA you're looking for? I can provide a response by early next week.

vrapolinario (commented):
Review of Flannel portion. Will review Calico later.

guides/guide-for-adding-windows-node.md
```diff
@@ -1,229 +1,402 @@
...
-
 ```
 
 2. Download & configure Flannel for Linux
 
 Download the most recent Flannel manifest:
 
 ```bash
-
 wget https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml
```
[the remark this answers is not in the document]
fabi200123: These will be run on Linux. After modifying it as James suggested, and as it writes above, I think it's pretty self explanatory that it should be run on Linux environment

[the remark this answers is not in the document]
fabi200123: I will also add a Note at the begining, after the "Configuring Flannel..." to tell people that they need to run those commands on a Linux environment

guides/guide-for-adding-windows-node.md
```diff
@@ -1,229 +1,402 @@
...
+# Install wins, kubelet and kubeadm
+.\PrepareNode.ps1 -KubernetesVersion v1.24.3 -ContainerRuntime containerD
 ```
 
-After a few minutes, you should see all the pods as running if the Flannel pod network was deployed.
+>  **Note** If you want to install another version of kubernetes, modify v1.24.3 with the version you want to install
+
+
+2. Run `kubeadm` to join the node
+>  **Note** Before joining the node, copy the file from /run/flannel/subnet.env to your windows machine to C:\run\flannel\subnet.env
```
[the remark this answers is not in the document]
fabi200123: This one is a little bit tricky. That file is being created on the Linux machine and people need to copy it to their windows machine at that specific path. If they don't do this, then it won't work properly and run into some problems with the flannel pods

guides/guide-for-adding-windows-node.md
fabi200123: I will update it to be more generic. There is also a flannel hostprocess section and wanted to differentiate them

guides/guide-for-adding-windows-node.md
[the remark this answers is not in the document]
vrapolinario: Got it. I think it's reasonable to expect that someone adding a Windows node to the scenario would be more comfortable using a Windows PS session. My suggestion here is to add the instructions to run these commands in PS - it's not much of a change anyway and would be greatly appreciated by Windows users.

guides/guide-for-adding-windows-node.md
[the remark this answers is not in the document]
vrapolinario: I see. Thanks for clarifying. It is tricky and just wondering if a Windows user would know how to perform this - but no further comments on the doc itself.

guides/guide-for-adding-windows-node.md
```diff
@@ -1,229 +1,402 @@
...
+>  **Note** If you want to install another version of kubernetes, modify v1.24.3 with the version you want to install
+
+
+2. Run `kubeadm` to join the node
+
+Use the command that was given to you when you ran `kubeadm init` on a control plane host. If you no longer have this command, or the token has expired, you can run `kubeadm token create --print-join-command` (on a control plane host) to generate a new token and join command.
+
+>  **Note:** Do not forget to add `--cri-socket "npipe:////./pipe/containerd-containerd" --v=5` at the end of the join command, if you use ContainerD
+
+3. Install kubectl for windows (optional)
```
lippertmarkus, line 250: Use "Windows" consistently throughout the doc

guides/guide-for-adding-windows-node.md
```diff
@@ -1,229 +1,402 @@
...
 
+```shell
+kubectl -n kube-system get pods -l app=flannel
 ```
 
-The output should include the Linux flannel DaemonSet as running:
+Once the flannel Pod is running, your node should enter the `Ready` state and then be available to handle workloads.
+
+
+## Configuring Calico hostprocess
```
lippertmarkus, line 160: Add note about required Kubernetes/Windows version for hostprocess containers to be available

guides/guide-for-adding-windows-node.md
```diff
@@ -1,229 +1,402 @@
...
+5. Apply kube-calico-rbac.yml from sig-windows-tools/kubeadm/flannel
+Next you will need to apply the configuration that allows calico to spawn pods and keep them running:
 
+```bash
+kubectl apply -f sig-windows-tools/hostprocess/calico/kube-calico-rbac.yml
 ```
 
->  **Note** If you are using another version of kubernetes on your windows node, change v1.24.3 with your own version .
 
+### Joining a Windows worker node
```
lippertmarkus, line 224: As the non-CNI-related steps on Linux and Windows as well as the verification steps are always the same, may think about consolidating them in a single paragraph instead of duplicating it multiple times

guides/guide-for-adding-windows-node.md
```diff
@@ -1,229 +1,402 @@
...
+Use the command that was given to you when you ran `kubeadm init` on a control plane host. If you no longer have this command, or the token has expired, you can run `kubeadm token create --print-join-command` (on a control plane host) to generate a new token and join command.
+
+>  **Note:** Do not forget to add `--cri-socket "npipe:////./pipe/containerd-containerd" --v=5` at the end of the join command, if you use ContainerD
+
+3. Install kubectl for windows (optional)
+
+For more information about it : https://kubernetes.io/docs/tasks/tools/install-kubectl-windows/
+
+
+#### Verifying your installation
```
lippertmarkus, line 255: As the non-CNI-related steps on Linux and Windows as well as the verification steps are always the same, may think about consolidating them in a single paragraph instead of duplicating it multiple times

lippertmarkus (commented):
Thanks for your effort! Generally I think there is too much duplication yet as we have many identical steps not depending on the used CNI. I would put those "generic" steps into seperate paragraphs and just differentiate for the CNI stuff. 

Also, as the CNI installation via hostprocess containers is IMHO much more easy and similar to the way to how it works on Linux I would center the docs more around that and add a "if you're below Kubernetes vX.Y.Z without hostprocess container support, you need to foollow the steps here" and link to a separate page.  This would make the whole document a bit more concise and less frightening.

guides/guide-for-adding-windows-node.md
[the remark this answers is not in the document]
fabi200123: Yes, it would but those are to be run before joining the node to the cluster, so for the moment all they have is their Linux master node. So I am not sure if I shall rewrite them for PS as well.

guides/guide-for-adding-windows-node.md
```diff
@@ -114,33 +269,32 @@ kubectl apply -f sig-windows-tools/kubeadm/flannel/kube-flannel-rbac.yml
 1. Install ContainerD, wins, kubelet, and kubeadm.
 
 ```PowerShell
-git clone https://github.com/kubernetes-sigs/sig-windows-tools
-cd .\sig-windows-tools\kubeadm\scripts\
-
 # Install ContainerD
```
lippertmarkus, line 272: should we use https://github.com/microsoft/Windows-Containers/blob/Main/helpful_tools/Install-ContainerdRuntime/install-containerd-runtime.ps1 instead of the sig-windows scripts?

guides/guide-for-adding-windows-node.md
```diff
@@ -1,229 +1,402 @@
...
 
-> **Note:** The VNI must be set to 4096 and port 4789 for Flannel on Linux to interoperate with Flannel on Windows. See the [VXLAN documentation](https://github.com/coreos/flannel/blob/master/Documentation/backends.md#vxlan). for an explanation of these fields.
+4. Add Windows Flannel and kube-proxy DaemonSets
 
->  **Note:** To use L2Bridge/Host-gateway mode instead change the value of `Type` to `"host-gw"` and omit `VNI` and `Port`.
+Now you can add Windows-compatible versions of calico and kube-proxy. In order to ensure that you get a compatible version of kube-proxy, you'll need to substitute the tag of the image. The following example shows usage for Kubernetes v1.24.3, but you should adjust the version for your own deployment.
 
-3. Apply the Flannel manifest and validate
+```bash
+curl -L https://github.com/kubernetes-sigs/sig-windows-tools/releases/latest/download/kube-proxy.yml | sed 's/VERSION/v1.24.3/g' | kubectl apply -f -
```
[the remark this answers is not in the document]
fabi200123: As I've said above, I am not sure if this part shall be written to run on a PowerShell session.

guides/guide-for-adding-windows-node.md
vrapolinario: I agree that leveraging the official scripts to deploy containerd is ideal. Plus, I had a comment above on which my recommendation is to stay away from cloning entire repos on the user environment. Totally fine to pull a specific file/script, but pulling a repo has two downsides: 1) It might leave unwanted files on the user environment and 2) requires the user to use GIT which is not a requirement to use Kubernetes.

jsturtevant: I like the idea of having a one way of installing containerd. It will avoid confusion in the long run. The install script on the MS repo is new but looks like we could leverage it (haven't tried it myself but it was based off a blog post I did).  The install-containerd script in this repo was a convivence since at the time there wasn't a ton of detail on how to do it. Since we now have the one from MS we could leverage that.  

Two concerns:
- being able to get changes from community merged quickly. I think myself and @marosset can help with that
- It installs nerdctl which isn't ready for more than basic commands and might cause more issues for folks that try to use it 

Maybe we can submit a pr to the MS script to optionally turn off nerdctl, which would test concern number one :-)

Note if we go that route, we should probably mark the one in this repo as deprecated in some way (there are some know issues with it anyways).  We also need to ensure that we don't install the CNI using the MS script.

@marosset thoughts?

lippertmarkus: I also talked to @brasmith-ms that I generally would advocate to reduce the number of scripts for installation. We currently have a ton of different ones, and it just causes confusion on which is the newest one, etc.

I also talked to him about having a more clean installation path. As WinGet doesn't have support for Windows Server, we could also start with chocolatey packages. I would dream of something more similar to the linux way:

```powershell
# install container runtime and Windows service
choco install containerd
# install kubelet as a Windows service
choco install kubelet kubeadm kubectl

# ... kubeadm join on Windows node would automatically configure kubelet service  ...
# ... (kubeadm init on Linux node already installed hostprocess kube-proxy Windows pod) ...
# ... now we could install CNI plugin via hostprocess container by kubectl apply ...
```
Could we work on something like that instead of the scripts? I would be happy to help.

guides/guide-for-adding-windows-node.md
```diff
@@ -25,9 +24,16 @@ Your Kubernetes server must be at or later than version 1.22. To check the versi
...
+- flannel
+- calico hostprocess
+- flannel hostprocess
+
+>  **Note:** You will need to use only one of them.
 
+## Configuring Flannel on Linux nodes
 
-#### Configuring Flannel with rancher
+>  **Note:** All code snippets in Linux sections are to be run in a Linux environment on the Linux worker node.
```
claudiubelu, line 37: Hm, moving forward, I think the hostprocess deployments would be the way to go. Thus, I'd interchange ``flannel`` with ``flannel hostprocess`` sections. I think that could lead to less confusion as well. I was following this guide, going to deploy ``flannel hostprocess``, but ended up following this section as well as I thought this was needed as well, including the Windows bits below, without noticing that the Windows bits below are not for ``flannel hostprocess``.

Or maybe even group up the flannel guide together, since it seems that the Linux side is identical on both cases. The main difference I see is the yamls that have to be applied.

fabi200123: Updated it in order to use only flannel hostprocess. I believe this will make it easier to follow and less confusing.

ameade:
Ty, this helped me get Flannel and Containerd working together on k8s 1.25. You might want to note that containerd must be at least version 1.6 for host processes to work. I hit this since the Install-Containerd.ps1 script defaulted to 1.4

fabi200123:
Updated the script to install containerd v1.6.8 and changed in the guide that you also need containerd version over 1.6. Thank you very much @ameade

guides/guide-for-adding-windows-node.md
brasmith-ms: I'm fine with either option, in fact if the chocolatey package turns out to be a valid option I can update the script in the MS repo to use this instead. The WindowsContainers repo is under the control of the Windows container product group at MS so I am not worried about that disappearing as long as Windows containers exist :)

jsturtevant: I don't think it should be a requirement for sys admins to install Chocolatey or scoop on their servers to set up Windows.  Anyone with that type of knowledge should be able to figure out how to tweak the deployment scripts to use those tools.  

When and if Winget is officially supported in WS I would be in favor of using.

fabi200123:
@marosset @jsturtevant I think we should try and add the modifed version of the guide. On this repository it is not the updated guide (or at least it does not work anymore for 1.25+)

hostprocess/flannel/kube-proxy/kube-proxy.yml
```diff
@@ -21,14 +21,16 @@ spec:
...
+      - image: sigwindowstools/kube-proxy:v1.24.3-flannel-hostprocess
         name: kube-proxy
         imagePullPolicy: Always
         env:
         # kube-proxy will use this value to select the correct network
         # https://github.com/kubernetes/kubernetes/blob/b0bc8adbc2178e15872f9ef040355c51c45d04bb/pkg/proxy/winkernel/proxier.go#L155-L158
         - name: KUBE_NETWORK
           value: "flannel.4096"
+        - name: CNI_BIN_PATH
+          value: C:\\Program Files\\containerd\\cni\\bin
```
jsturtevant, line 33: This path doesn't match with the path that the `Install-Containerd.ps1` path:

https://github.com/kubernetes-sigs/sig-windows-tools/blob/f366740c551b4217887a9eb339fd7ec3d6caacce/kubeadm/scripts/Install-Containerd.ps1#L123-L124

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -67,7 +67,7 @@ $env:Path += ";$global:KubernetesPath"
 
 DownloadFile $kubeletBinPath https://dl.k8s.io/$KubernetesVersion/bin/windows/amd64/kubelet.exe
 DownloadFile "$global:KubernetesPath\kubeadm.exe" https://dl.k8s.io/$KubernetesVersion/bin/windows/amd64/kubeadm.exe
-DownloadFile "$global:KubernetesPath\wins.exe" https://github.com/rancher/wins/releases/download/v0.0.4/wins.exe
+DownloadFile "$global:KubernetesPath\wins.exe" https://github.com/rancher/wins/releases/download/v0.4.11/wins.exe
```
jsturtevant, line 70: We shouldn't need wins anymore with the Hostprocess approach

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -100,7 +100,9 @@ $cmd_commands=@("C:\k\kubelet.exe ", '$global:KubeletArgs ', '--cert-dir=$env:SY
 if ($CurrentVersion -lt $V1_24_Version) {
     $cmd_commands = $cmd_commands + "--network-plugin=cni " + "--image-pull-progress-deadline=20m "
 }
-$StartKubeletFileContent = '$FileContent = Get-Content -Path "/var/lib/kubelet/kubeadm-flags.env"
+$StartKubeletFileContent = 'Import-Module "c:\k\hns.psm1"
+New-HnsNetwork -Type NAT -Name nat
```
jsturtevant, line 104: we shouldn't need to create the NAT network with Hostprocess approach.

jsturtevant (changes requested):
We don't need to create the NAT config for containerd with hostprocess in the `Install-container.ps1` script. This may cause issues with network switching between NAT and the calico/flannel networks that are created.
https://github.com/kubernetes-sigs/sig-windows-tools/blob/f366740c551b4217887a9eb339fd7ec3d6caacce/kubeadm/scripts/Install-Containerd.ps1#L140-L159

kubeadm/scripts/PrepareNode.ps1
fabi200123: Bumped into an issue and from what I read, this was a solution to it. I will modify it back

hostprocess/calico/kube-calico-rbac.yml
```diff
@@ -0,0 +1,95 @@
+---
+apiVersion: v1
```
jsturtevant, line 2: should these roles already be present from the Linux calico deployment?

fabi200123: From what I tested, they weren't and needed to be applied separately

jsturtevant: ah, I think this might be due to the way calico is installed here. https://projectcalico.docs.tigera.io/manifests/tigera-operator.yaml vs https://github.com/projectcalico/calico/releases/download/$(CALICO_VERSION)/release-$(CALICO_VERSION).tgz 

I used the released files when developing the files in the calico folder. I bet if those are used we wouln't need this. 

I don't really know the difference between the release files and operator. do you?

fabi200123: Unfortunatelly, I don't know either

jsturtevant: Given the object naming is different, I think it makes sense to use the files in to be consistent  https://github.com/projectcalico/calico/releases/download/$(CALICO_VERSION)/release-$(CALICO_VERSION).tgz. Otherwise we end up with two roles that are basically the same for just named different because of the operator. 

btw, thanks for all the hard work on this. I am testing the changes locally but things look lgtm

hostprocess/flannel/kube-proxy/kube-proxy.yml
```diff
@@ -21,14 +21,16 @@ spec:
...
+      - image: sigwindowstools/kube-proxy:v1.24.3-flannel-hostprocess
         name: kube-proxy
         imagePullPolicy: Always
         env:
         # kube-proxy will use this value to select the correct network
         # https://github.com/kubernetes/kubernetes/blob/b0bc8adbc2178e15872f9ef040355c51c45d04bb/pkg/proxy/winkernel/proxier.go#L155-L158
         - name: KUBE_NETWORK
           value: "flannel.4096"
+        - name: CNI_BIN_PATH
+          value: C:\opt\cni\bin
```
jsturtevant, line 33: this isn't the same as the flannel install path above (C:\\Program Files\\containerd\\cni\\bin)

jsturtevant: since the script uses the default containerd paths, it will need to be `C:\Program Files\containerd\cni\bin`

guides/guide-for-adding-windows-node.md
```diff
@@ -90,20 +95,88 @@ kube-system kube-flannel-ds-54954 1/1 Running 0 1m
...
 > To find your version of kubernetes run the following command:
 > `kubeadm version`
 
 5. Apply kube-flannel-rbac.yml from sig-windows-tools/kubeadm/flannel
 Next you will need to apply the configuration that allows flannel to spawn pods and keep them running:
 
 ```bash
-git clone https://github.com/kubernetes-sigs/sig-windows-tools
-kubectl apply -f sig-windows-tools/kubeadm/flannel/kube-flannel-rbac.yml
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/sig-windows-tools/master/kubeadm/flannel/kube-flannel-rbac.yml
```
jsturtevant, line 113: I think this shouldn't be needed.  Looks like flannel .17 manifests moved all the components to `kube-flannel` namespace.  We should update our manifests to delpoy to `kube-flannel` instead of `kube-system`, which that update we shouldn't need any additional RBAC

guides/guide-for-adding-windows-node.md
```diff
@@ -90,20 +95,88 @@ kube-system kube-flannel-ds-54954 1/1 Running 0 1m
 Now you can add Windows-compatible versions of Flannel and kube-proxy. In order to ensure that you get a compatible version of kube-proxy, you'll need to substitute the tag of the image. The following example shows usage for Kubernetes v1.24.3, but you should adjust the version for your own deployment.
 
 ```bash
-curl -L https://github.com/kubernetes-sigs/sig-windows-tools/releases/latest/download/kube-proxy.yml | sed 's/VERSION/v1.24.3/g' | kubectl apply -f -
-kubectl apply -f https://github.com/kubernetes-sigs/sig-windows-tools/releases/latest/download/flannel-overlay.yml
+# Make sure to modify the version of kube-proxy image to match your kubernetes version
+# before applying it
+wget https://raw.githubusercontent.com/kubernetes-sigs/sig-windows-tools/master/hostprocess/flannel/kube-proxy/kube-proxy.yml
+kubectl apply -f sig-windows-tools/hostprocess/flannel/kube-proxy/kube-proxy.yml
```
jsturtevant, line 101: wget downloads this file as `kube-proxy.yml` not the full path.  We should apply that file or use the SED trick used in the original.

guides/guide-for-adding-windows-node.md
```diff
@@ -114,33 +187,32 @@ kubectl apply -f sig-windows-tools/kubeadm/flannel/kube-flannel-rbac.yml
...
 
 ```PowerShell
-git clone https://github.com/kubernetes-sigs/sig-windows-tools
-cd .\sig-windows-tools\kubeadm\scripts\
-
 # Install ContainerD
+curl.exe -LO https://raw.githubusercontent.com/kubernetes-sigs/sig-windows-tools/master/kubeadm/scripts/Install-Containerd.ps1
 .\Install-Containerd.ps1
 
 # Install wins, kubelet and kubeadm
```
jsturtevant, line 194: We don't need wins any more

jsturtevant: should remove references from the guide so not to confuse

hostprocess/flannel/kube-proxy/kube-proxy.yml
jsturtevant: I just noticed the Install-containerd.ps1 script does update the containerd config to `C:\opt\cni\bin`. It doesn't really matter what the folders are, but we do need to be consistent across these files.

kubeadm/scripts/Install-Containerd.ps1
```diff
@@ -137,27 +137,6 @@ $prefixLength = (Get-NetIPAddress -InterfaceAlias $netAdapterName -AddressFamily
 
 $subnet = CalculateSubNet -gateway $gateway -prefixLength $prefixLength
 
-@"
```
jsturtevant, line 140: we should clean up the line above this as well: 

```
Write-Output "Creating network config for nat network"
$gateway = (Get-NetIPAddress -InterfaceAlias $netAdapterName -AddressFamily IPv4).IPAddress
$prefixLength = (Get-NetIPAddress -InterfaceAlias $netAdapterName -AddressFamily IPv4).PrefixLength
```

hostprocess/flannel/kube-proxy/kube-proxy.yml
jsturtevant: the script also downloads some cni binaries and sticks them into `C:\opt\cni\bin`. We don't need that anymore either as the CNI are set up to bring their own required binaries.

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -82,10 +81,6 @@ if ($ContainerRuntime -eq "Docker") {
     New-HnsNetwork -Type NAT -Name nat
```
jsturtevant, line 81: We shouldn't create this, it could lead to strange networking issues.

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -67,7 +67,6 @@ $env:Path += ";$global:KubernetesPath"
 
 DownloadFile $kubeletBinPath https://dl.k8s.io/$KubernetesVersion/bin/windows/amd64/kubelet.exe
 DownloadFile "$global:KubernetesPath\kubeadm.exe" https://dl.k8s.io/$KubernetesVersion/bin/windows/amd64/kubeadm.exe
-DownloadFile "$global:KubernetesPath\wins.exe" https://github.com/rancher/wins/releases/download/v0.0.4/wins.exe
 
 if ($ContainerRuntime -eq "Docker") {
```
jsturtevant, line 71: We could likely just remove all references to docker here.  It will not work with hostprocess containers and isn't support in 1.24+

hostprocess/flannel/flanneld/flannel-overlay.yml
```diff
@@ -111,6 +111,10 @@ spec:
         - name: kube-proxy
           mountPath: /flannel-config-file
```
jsturtevant, line 112: The other issue here is that if flannel has moved to the `kube-flannel` ns and we update our deployments to do that then the kubeproxy and other kubeadm configs we were relying on can't be accessed since they live in kube-system. Will need to maybe create our configmaps or change the way those files are used.

jsturtevant: I looked into this a bit, The kubeproxy cm is used for the flannel connection to the api-server instead of using that we should use the token that get projected into the account. With containerd 1.6 if still requires so changes due to `CONTAINER_SANDBOX_MOUNT_POINT` and mounting but we can add a new cm to the deployment files that could be used. In the next version of containerd we can use in the cluster config that flannel provides

The kubeadm cm is used to get the pod and service cidrs.  This information pod cidr is already in `kube-flannel-cfg` in the ns `kube-flannel` in 0.17.  The service cider is something is static so we could just move it to the `kube-flannel-cfg-windows` that we create and read it from there.

This would remove the for the kube-adm and kube-proxy configmaps which out having to add the addtional RBAC rules to allow reading of those files.

jsturtevant (changes requested):
I went through this manually and found a few things that need to be resolved for flannel.  Left comments where changes are necessary. Let me know if I can clarify anything.

hostprocess/flannel/kube-proxy/kube-proxy.yml
fabi200123: updated

guides/guide-for-adding-windows-node.md
fabi200123: done

kubeadm/scripts/PrepareNode.ps1
fabi200123: removed this

kubeadm/scripts/PrepareNode.ps1
fabi200123: I removed all references to docker. It is only containerD now

hostprocess/flannel/kube-proxy/kube-proxy.yml
fabi200123: I modified the path

hostprocess/calico/kube-calico-rbac.yml
fabi200123: Updated the guide to use calico from projectcalico! If there is something that I might have missed or can be modified, please let me know!

kubeadm/scripts/Install-Containerd.ps1
meilel: why do we still calculat the subnet if we don't need the nat config anymore below?
Am I missing something or is this code obsolet?

kubeadm/scripts/Install-Containerd.ps1
```diff
@@ -120,48 +120,24 @@ $env:Path += ";$global:ConainterDPath"
...
+$config = $config -replace "conf_dir = (.)*$", "conf_dir = `"C:\\Program Files\\containerd\\cni\\conf`""
 $config | Set-Content "$global:ConainterDPath\config.toml" -Force 
 
-mkdir -Force c:\opt\cni\bin | Out-Null
-mkdir -Force c:\etc\cni\net.d | Out-Null
-
 Write-Output "Getting SDN CNI binaries"
-DownloadFile "c:\opt\cni\cni-plugins.zip" https://github.com/microsoft/windows-container-networking/releases/download/v0.2.0/windows-container-networking-cni-amd64-v0.2.0.zip
-Expand-Archive -Path "c:\opt\cni\cni-plugins.zip" -DestinationPath "c:\opt\cni\bin" -Force
+DownloadFile "$global:ConainterDPath\cni\cni-plugins.zip" https://github.com/microsoft/windows-container-networking/releases/download/v0.2.0/windows-container-networking-cni-amd64-v0.2.0.zip
```
meilel, line 128: We should create the folder $global:ConainterDPath\cni\ first.
Because I get an error if the folder does not exist when dowloading the cni-plugins.zip

guides/guide-for-adding-windows-node.md
```diff
@@ -90,57 +95,112 @@ kube-system kube-flannel-ds-54954 1/1 Running 0 1m
...
+```
+
+
+## Configuring Calico hostprocess
+
+>  **Note:** All code snippets in Linux sections are to be run in a Linux environment on the Linux worker node.
+
+1. Prepare Kubernetes control plane for Calico
+
+Some minor preparation is recommended on the Kubernetes control plane in our cluster. It is recommended to enable bridged IPv4 traffic to iptables chains when using Flannel. The following command must be run on all Linux nodes:
```
mweibel, line 122: `when using Flannel` in the Calico topic

fabi200123: Forgot to modify it to calico from flannel

Mik4sa:
@fabi200123 I closely watched your PRs in the last days and in general was able to get by windows node (Windows Server 2022) "registered". The pods weren't failing and `kubectl get nodes` showed the node as "ready". BUT I wasn't able to schedule a pod on that machine. I got multiple errors.
The first erorr that occured was: `incompatible CNI versions; config is "0.3.1", plugin supports ["0.2.0"]`

I was able to fix all problems I found and I created a separate fork/branch for that.

Maybe you would like to have a look at some changes, maybe you need some of them too to finish this PR.
https://github.com/kubernetes-sigs/sig-windows-tools/compare/master...Mik4sa:sig-windows-tools:flannel-hostprocess

Don't hestitate to contact me in case you need some more info from me.

Note: I'm still faily new and I'm not really sure what I did there. Everything was trial and error.

fabi200123:
@Mik4sa Thank you for your work! I will look over it soon. Also, there are some new PRs open to solve some issues. The temporary guide PR was already merged, but the [one updating the scripts](https://github.com/kubernetes-sigs/sig-windows-tools/pull/262) is still open. Try to check that one out and the new guide as well. And also, be careful what CNI you are using and that you have that specific cni on the windows node as well. Also, I believe the flannel 0.14 version is a little bit old and I should also change it to 0.17 or a newer one. Try with flannel v0.17 hostprocess. Let me know if this helped or I shall explain more.

Mik4sa:
Why not directly use the latest version 0.20.2 then? Are there any breaking changes you know of?
Though I'll stay for flannel 0.14.0 for now. It seems to work in general and I would like to test my use case with my windows cluster before testing other stuff
Thank you :)

fabi200123:
> Why not directly use the latest version 0.20.2 then? Are there any breaking changes you know of? Though I'll stay for flannel 0.14.0 for now. It seems to work in general and I would like to test my use case with my windows cluster before testing other stuff Thank you :)

Didn't checked for the last version of flannel to see if it supports config "0.3.1", but from what I tested, v0.17 did support that one.
Also, let me know if you need help with anything else.

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -138,11 +99,7 @@ $newPath = "$global:NssmInstallDirectory;" +
 Write-Host "Registering kubelet service"
 nssm install kubelet $global:Powershell $global:PowershellArgs $global:StartKubeletScript
 
-if ($ContainerRuntime -eq "Docker") {
-    nssm set kubelet DependOnService docker
-} elseif ($ContainerRuntime -eq "containerD") {
-    nssm set kubelet DependOnService containerd
-}
+nssm set kubelet DependOnService containerd
```
doctorpangloss, line 102: The kubelet service script needs to be aware of which networking it is using due to Calico issues, see https://github.com/projectcalico/calico/pull/6656/files

PR #240, opened 2022-08-12
Updating host-process images to use the new base image

marosset:
Signed-off-by: Mark Rossetti <marosset@microsoft.com>

**Reason for PR**:

Updating host-process images to use the new base image

**Issue Fixed**:

Issue #

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests

**Notes**:

### DIFF
--- a/hostprocess/calico/install/Dockerfile.install
+++ b/hostprocess/calico/install/Dockerfile.install
@@ -1,7 +1,4 @@
-# Note this image doesn't really mater for hostprocess
-# the files in the image are copied to $env:CONTAINER_SANDBOX_MOUNT_POINT on the host
-# but the file system is the Host NOT the container
-ARG BASE="mcr.microsoft.com/windows/nanoserver:1809"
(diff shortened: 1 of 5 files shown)


hostprocess/eventflow-logger/Dockerfile
```diff
@@ -1,5 +1,4 @@
-ARG OS=1809
-FROM mcr.microsoft.com/windows/nanoserver:${OS} as runtime
+FROM mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v0.1.0 as runtime
```
jsturtevant, line 1: this image is currently using Docker not buildx so I think it won't work. Will need to update the build scripts

marosset: Yea, I noticed taht. Same w/ csi-proxy.
I might just update the images that are building w/ buildkit in the PR and then open a separate PR for these 2 images.

PR #241, opened 2022-08-17
Adding hack/sync-project-board.sh

marosset:
**Reason for PR**:

Adding a script that will look at all repos in https://github.com/kubernetes for issues with `sig/windows` label and add them to https://github.com/orgs/kubernetes/projects/82

**Notes**:
 it is not currently possible to add items from other orgs to a 'new' project via github graphql mutations.

We can maybe set up another project board for kuberentes-sigs and do something similar.

**Issue Fixed**:

Issue #

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/hack/sync-project-board.sh
+++ b/hack/sync-project-board.sh
@@ -1,0 +1,83 @@
+#!/bin/bash
+set -e
+set -u
+set -o pipefail
+
+GH_ORG=${GH_ORG:-'kubernetes'}
+PROJECT_NUMBER=${PROJECT_NUMBER:-'82'}
+
+echo "GH_ORG=${GH_ORG}"
+
+# Get project ID
+project_id="$(gh api graphql -f query='
+    query($org: String!, $number: Int!) {
+        organization(login: $org) {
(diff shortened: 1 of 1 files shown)


marosset:
Next I can look into running this as a GH action

jsturtevant:
can the script be run multiple times? If you move things around in the project board (new view or change add a new field) does that stick?

marosset:
> can the script be run multiple times? If you move things around in the project board (new view or change add a new field) does that stick?

Yes to all of the above.
If you call `addProjectV2ItemById` and add an item already on the project board it just returns the id (or whatever other information you request in the query). This is intentional and was recommended behavior in the docs.

This same call only adds an item to the project board and cannot be used to change field values for the item (including position on the board). 
There are other mutations for doing those (`UpdateProjectV2ItemFieldValue`, `UpdateProjectV2ItemPosition`, etc)

jsturtevant:
/lgtm
we can iterate as we use it more

PR #247, opened 2022-09-20
Solving the issue for  K8s 1.25

fabi200123:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it:**
If you apply the windows kube-proxy field directly it fails. 
This PR is based on @jsturtevant [PR](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2488) to solve the same issue on this repository.


marosset:
Can we have more descriptive PR titles in the future? :)

PR #248, opened 2022-10-26
Scripts update to make them usable with hostprocess model

a4099181:
**Reason for PR**:
It seems that `Install-Container.ps1` and `PrepareNode.ps1` may be useful to build a Windows node with HostProcess containers and k8s 1.25.3 at least. But some changes are needed because some code does not cooperate. This commit is a proposal to leave the current code unchanged for backward compatibility and introduce the _UseHostProcess_ parameter to handle code not needed to support HostProcess containers where are engaged.

**Requirements**

- [x] Squash commits 

**Notes**:
There are also changes beyond the main reason. When the UseHostProcess flag is present, the script makes asserts on Kubernetes and ContainerD versions to announce incompatibilities. Another improvement is to not have a choice of container runtime when HostProcess is in use.

Additionally, it offers the _SuppressHints_ parameter to have the option to omit on demand some help messages. Omitted code depends on the git tool installed. It may be an optimization for automated node provisioning (eg. vagrant) and saves some time wasted for git install and repository cloning.

### DIFF
--- a/kubeadm/scripts/Install-Containerd.ps1
+++ b/kubeadm/scripts/Install-Containerd.ps1
@@ -11,6 +11,7 @@ This script
 
 .PARAMETER ContainerDVersion
 ContainerD version to download and use.
+For HostProcess containers minimal version is 1.6.0.
 
 .PARAMETER netAdapterName
 Name of network adapter to use when configuring basic nat network.
@@ -20,6 +21,9 @@ Declares that HostProcess model is in use. Some code should be omitted.
 
 .EXAMPLE
 PS> .\Install-Conatinerd.ps1
+
+.EXAMPLE
+PS> .\Install-Containerd.ps1 -UseHostProcess -ContainerDVersion # 1.6.0+
 
 #>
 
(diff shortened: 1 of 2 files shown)


marosset:
The behavior of volume mounts changes depending on if you are using containerd `v1.6.x` vs `v1.7.x` (more info at https://github.com/kubernetes/enhancements/tree/master/keps/sig-windows/1981-windows-privileged-container-support#container-mounts)

Can we update all the references to default to https://github.com/containerd/containerd/releases/tag/v1.7.0-beta.1 instead of v1.6.8 (until v1.7.0 releases)?

kubeadm/scripts/Install-Containerd.ps1
```diff
@@ -11,20 +11,38 @@ This script
...
 Param(
     [parameter(HelpMessage = "ContainerD version to use")]
     [string] $ContainerDVersion = "1.6.8",
-    [parameter(HelpMessage = "Name of network adapter to use when configuring basic nat network")]
-    [string] $netAdapterName = "Ethernet"
+    [parameter(
+        HelpMessage = "Name of network adapter to use when configuring basic nat network",
+        ParameterSetName = "kubeadm"
+    )]
+    [string] $netAdapterName = "Ethernet",
```
marosset, line 40: I think we can drop the `$netAdapterName` param since this script no longer creates a nat network (removed with https://github.com/kubernetes-sigs/sig-windows-tools/pull/254 b/c it was causing lots of issues)

marosset (changes requested):
Left a couple of comments

kubeadm/scripts/Install-Containerd.ps1
a4099181: Yes, now it is unused parameter. Should we take care of the backward compatibility? If not, then we can do some more cleanups in both scripts. If so, these scripts will not fit kubeadm folder no longer. Maybe better/cleaner version of the scripts should be made in host-process folder. But there is a couple of PR's which makes changes to them right now. It is not a good time, I think.

I am not convinced to update referenced containerd version in this PR.

PR #262, opened 2022-12-20
Updating Windows Scripts

fabi200123:
**Reason for PR**:
To update the Windows scripts: InstallContainerd.ps1 and PrepareNode.ps1. 
The location for containerd was changed to the correct one in the script and the script which installs kubeadm, kubelet was changed to install k8s version v1.25.3 and Docker references were removed because they are deprecated!

**Issue Fixed**:
Problems which may appear because containerd folders are not properly set in the script.

**Notes**:
This PR is based on the same changes to the scripts as PR [#239](https://github.com/kubernetes-sigs/sig-windows-tools/pull/239)

### DIFF
--- a/kubeadm/scripts/Install-Containerd.ps1
+++ b/kubeadm/scripts/Install-Containerd.ps1
@@ -81,17 +81,10 @@ $config = $config -replace "bin_dir = (.)*$", "bin_dir = `"c:/opt/cni/bin`""
 $config = $config -replace "conf_dir = (.)*$", "conf_dir = `"c:/etc/cni/net.d`""
 $config | Set-Content "$global:ConainterDPath\config.toml" -Force 
 
-mkdir -Force c:\opt\cni\bin | Out-Null
-mkdir -Force c:\etc\cni\net.d | Out-Null
-
-Write-Output "Getting SDN CNI binaries"
-DownloadFile "c:\opt\cni\cni-plugins.zip" https://github.com/microsoft/windows-container-networking/releases/download/v0.2.0/windows-container-networking-cni-amd64-v0.2.0.zip
-Expand-Archive -Path "c:\opt\cni\cni-plugins.zip" -DestinationPath "c:\opt\cni\bin" -Force
-
 Write-Output "Registering ContainerD as a service"
 containerd.exe --register-service
 
 Write-Output "Starting ContainerD service"
 Start-Service containerd
 
-Write-Output "Done - please remember to add '--cri-socket `"npipe:////./pipe/containerd-containerd`"' to your kubeadm join command"
+Write-Output "Done - please remember to add '--cri-socket `"npipe:////./pipe/containerd-containerd`"' to your kubeadm join command if your kubernetes version is below 1.25!"
--- a/kubeadm/scripts/PrepareNode.ps1
+++ b/kubeadm/scripts/PrepareNode.ps1
@@ -11,20 +11,14 @@ This script assists with joining a Windows node to a cluster.
 .PARAMETER KubernetesVersion
 Kubernetes version to download and use
 
-.PARAMETER ContainerRuntime
-Container that Kubernetes will use. (Docker or containerD)
-
 .EXAMPLE
-PS> .\PrepareNode.ps1 -KubernetesVersion v1.24.2 -ContainerRuntime containerD
+PS> .\PrepareNode.ps1 -KubernetesVersion v1.25.3
 
 #>
 
 Param(
     [parameter(Mandatory = $true, HelpMessage="Kubernetes version to use")]
-    [string] $KubernetesVersion,
-    [parameter(HelpMessage="Container runtime that Kubernets will use")]
-    [ValidateSet("containerD", "Docker")]
-    [string] $ContainerRuntime = "Docker"
+    [string] $KubernetesVersion
 )
 $ErrorActionPreference = 'Stop'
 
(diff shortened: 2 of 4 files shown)


kubeadm/scripts/Install-Containerd.ps1
```diff
@@ -77,21 +77,19 @@ $env:Path += ";$global:ConainterDPath"
...
 $config | Set-Content "$global:ConainterDPath\config.toml" -Force 
 
-mkdir -Force c:\opt\cni\bin | Out-Null
-mkdir -Force c:\etc\cni\net.d | Out-Null
-
 Write-Output "Getting SDN CNI binaries"
-DownloadFile "c:\opt\cni\cni-plugins.zip" https://github.com/microsoft/windows-container-networking/releases/download/v0.2.0/windows-container-networking-cni-amd64-v0.2.0.zip
-Expand-Archive -Path "c:\opt\cni\cni-plugins.zip" -DestinationPath "c:\opt\cni\bin" -Force
+mkdir -Force $global:ConainterDPath\cni | Out-Null
+DownloadFile "$global:ConainterDPath\cni\cni-plugins.zip" https://github.com/microsoft/windows-container-networking/releases/download/v0.2.0/windows-container-networking-cni-amd64-v0.2.0.zip
```
jsturtevant, line 86: With the hostprocess images, these binaries are ship as part of the installer for the image, we shouldn't need to install them at all.

kubeadm/scripts/Install-Containerd.ps1
```diff
@@ -77,21 +77,19 @@ $env:Path += ";$global:ConainterDPath"
...
 
 Write-Output "Registering ContainerD as a service"
 containerd.exe --register-service
 
 Write-Output "Starting ContainerD service"
 Start-Service containerd
 
-Write-Output "Done - please remember to add '--cri-socket `"npipe:////./pipe/containerd-containerd`"' to your kubeadm join command"
\ No newline at end of file
+Write-Output "Done - please remember to add '--cri-socket `"npipe:////./pipe/containerd-containerd`"' to your kubeadm join command"
```
jsturtevant, line 95: fyi - This shouldn't be needed 1.25+.  Kubeadm automatically detects this after 1.25. It's ok to leave a note though.

kubeadm/scripts/Install-Containerd.ps1
```diff
@@ -77,21 +77,19 @@ $env:Path += ";$global:ConainterDPath"
 containerd.exe config default | Out-File "$global:ConainterDPath\config.toml" -Encoding ascii
 #config file fixups
 $config = Get-Content "$global:ConainterDPath\config.toml"
-$config = $config -replace "bin_dir = (.)*$", "bin_dir = `"c:/opt/cni/bin`""
-$config = $config -replace "conf_dir = (.)*$", "conf_dir = `"c:/etc/cni/net.d`""
+$config = $config -replace "bin_dir = (.)*$", "bin_dir = `"C:\\Program Files\\containerd\\cni\\bin`""
+$config = $config -replace "conf_dir = (.)*$", "conf_dir = `"C:\\Program Files\\containerd\\cni\\conf`""
```
jsturtevant, line 81: This doesn't align with the configuration for calico: https://github.com/kubernetes-sigs/sig-windows-tools/blob/6d113824ba7bda0b5c449883c4fc3ac97fa9ea56/hostprocess/calico/calico.yml#L164-L169

This will result in the wrong CNI configuration being used.  We need to align these two values.

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -67,24 +54,9 @@ $env:Path += ";$global:KubernetesPath"
...
-    Import-Module "c:\k\hns.psm1"
-    # TODO(marosset): check if network already exists before creatation
-    New-HnsNetwork -Type NAT -Name nat
-}
 
-Write-Host "Registering wins service"
-wins.exe srv app run --register
-start-service rancher-wins
+DownloadFile "c:\k\hns.psm1" https://github.com/Microsoft/SDN/raw/master/Kubernetes/windows/hns.psm1
+Import-Module "c:\k\hns.psm1"
```
jsturtevant, line 59: If this is needed by the CNI then it is included in the Host process CNI image, so not technically needed.

kubeadm/scripts/Install-Containerd.ps1
jsturtevant: https://github.com/kubernetes-sigs/sig-windows-tools/blob/6d113824ba7bda0b5c449883c4fc3ac97fa9ea56/hostprocess/calico/install/Dockerfile.install#L14-L15

https://github.com/kubernetes-sigs/sig-windows-tools/blob/6d113824ba7bda0b5c449883c4fc3ac97fa9ea56/hostprocess/flannel/flanneld/Dockerfile#L14-L19

jsturtevant (changes requested):
Thanks for breaking this out, much easier to review.  The most important part here is making sure the containerd config and the CNI config installers point to the same location.

kubeadm/scripts/Install-Containerd.ps1
fabi200123: Forgot to update the calico.yml as well

fabi200123: It is better to leave that as it is. Apparently, changing the containerd CNI location will result in a lot more problems to solve than just leaving it as it is. CNIs are configured to have that location as default and changing it will result in even more changes, for the moment it is better to leave it as it is.

jsturtevant: Makes sense to me to leave it as is.

PR #271, opened 2022-12-27
Remove dockershim related files

jsturtevant:
Signed-off-by: James Sturtevant <jstur@microsoft.com>

**Reason for PR**:

Removes files that no longer work now that dockershim has been removed.  Now that 1.23 is the latest version of Kubernetes supported we can use the HostProcess containers solution.  

Files prior to this can be found at https://github.com/kubernetes-sigs/sig-windows-tools/releases/tag/v0.1.6

**Issue Fixed**:

Issue #

**Requirements**

- [ ] Sqaush commits 
- [x] Documentation
- [ ] Tests

**Notes**:

/assign @marosset 
/hold


kubeadm/scripts/Install-Containerd.ps1
```diff
@@ -1,90 +0,0 @@
-<#
-.SYNOPSIS
-Installs ContainerD on a Windows machines in preperation for joining the node to a Kubernetes cluster.
```
fabi200123, line 3: Should we not move this script as well? I believe this is still needed to install ContainerD on Windows

jsturtevant: yes, good catch!

jsturtevant:
fyi @Mik4sa

kubeadm/scripts/Install-Containerd.ps1
jsturtevant: fixed

marosset (approved):
/lgtm

PR #280, opened 2023-01-20
Remove dependency on git from prepare node script

### LINKED ISSUE #278: kubeadm script PrepareNode.ps1: Remove requirement on git being installed
The `PrepareNode.ps1` script now seems to require git to be installed. For example on this line:

https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/kubeadm/scripts/PrepareNode.ps1#L113

This complicates the setup a lot since it is not easy to get git installed on Windows Server 2022 and also not easy to get it working in Powershell. Especially in an scripted automated way.

If possible, I would suggest to rework the parts of the script that require git to be installed so it does not need git.

jonaskello:
**Reason for PR**:
Eliminates dependency on git command

**Issue Fixed**:
278

**Requirements**

- [x] Sqaush commits 
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/hostprocess/Install-Containerd.ps1
+++ b/hostprocess/Install-Containerd.ps1
@@ -1,11 +1,11 @@
 <#
 .SYNOPSIS
-Installs ContainerD on a Windows machines in preperation for joining the node to a Kubernetes cluster.
+Installs ContainerD on a Windows machines in preparation for joining the node to a Kubernetes cluster.
 
 .DESCRIPTION
 This script
-- Verifies that Windows Features requried for running contianers are enabled (and enables then if they are not)
-- Downloads ContainerD binaries from from at the version specified.
+- Verifies that Windows Features requried for running containers are enabled (and enables them if they are not)
+- Downloads ContainerD binaries for the version specified.
 - Downloads Windows SND CNI plugins.
 - Registers ContainerD as a windows service.
 


jonaskello:
@jsturtevant I the gitpath in a variable and made it point to the latest version. It would be better if it pointed to the exact version but I don't see how that could be done without depending on cloning the repo and using git.

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -105,24 +105,26 @@ nssm set kubelet DependOnService containerd
...
 
+$gitpath='https://raw.githubusercontent.com/kubernetes-sigs/sig-windows-tools/master'
+
 Write-Output "Please remember that after you have joined the node to the cluster, that you have to apply the cni daemonset/service and the kube-proxy"
 Write-Output "Also remember that for kube-proxy you have to change the its version from the name of the image in the kube-proxy.yml to that of your kubernetes version `n"
 # rancher commands
 Write-Output "In case you use rancher, use the following commands:"
 Write-Output "For Windows you can use the following command: "
-Write-Output "(Get-Content `"$(git rev-parse --show-toplevel)/kubeadm/kube-proxy/kube-proxy.yml`") -Replace 'VERSION', '$KubernetesVersion' | Set-Content `"$(git rev-parse --show-toplevel)/kubeadm/kube-proxy/kube-proxy.yml`" `n"
+Write-Output "(Get-Content `"$gitpath/kubeadm/kube-proxy/kube-proxy.yml`") -Replace 'VERSION', '$KubernetesVersion' | Set-Content `"$gitpath/kubeadm/kube-proxy/kube-proxy.yml`" `n"
```
jsturtevant, line 115: is this right?

```
 Set-Content `"$gitpath/kubeadm/kube-
```

I would have thought this would be local path

kubeadm/scripts/PrepareNode.ps1
```diff
@@ -105,24 +105,26 @@ nssm set kubelet DependOnService containerd
 
 New-NetFirewallRule -Name kubelet -DisplayName 'kubelet' -Enabled True -Direction Inbound -Protocol TCP -Action Allow -LocalPort 10250
 
+$gitpath='https://raw.githubusercontent.com/kubernetes-sigs/sig-windows-tools/master'
```
jsturtevant, line 108: gitpath varaible is a bit misleading. maybe repositoryRoot or something else?

kubeadm/scripts/PrepareNode.ps1
jonaskello: Yes, I misunderstood the original intention here I think. I think this part of the script assumes that you git cloned the whole repo down. 

However the guide tells you to download this script with curl and run it. So in that case it will not run from within a cloned repo and local paths will not work. So for that reason I think the commands to apply the manifests for kubeproxy should refer to the online manifests by using http urls.

However then of course the `Get-Content` needs to be changed to `curl` commands. I'll do that and see if it makes more sense.

jonaskello: I ended up first downloading the manifest and then patching the local file with the version. Do you think this will work?

jonaskello:
Thinking some more on this, perhaps the last part of the script that prints instructions on how to setup kube-proxy could be removed and the instructions moved to the guide document instead?

claudiubelu:
LGTM, but needs a rebase.

jonaskello:
Should be rebased now

jsturtevant:
@jonaskello Could you squash the commits and make sure the commits are valid? (I am not really sure what is invalid about them but hoping squashing will get rid of the warning

jonaskello:
@jsturtevant Failed to fix this PR, making a new attempt in #299

PR #284, opened 2023-02-07
Updating PrepareNode.ps1

zarakM:
Unzipping command doesn't work without appending "-" with C flag.

**Reason for PR**:

**Issue Fixed**:

Issue #

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests

**Notes**:


jsturtevant:
@zarakM can you take a look at the CLA?

riverar:
Not seeing any issues here with `C` vs `-C` on linux or windows. Is there a repro case demonstrating an actual issue here?

jsturtevant:
I also haven't experienced any issues but not sure that it hurst to have the `-`.

/close
since we haven't heard back on the CLA, please re-open if still an issue.

PR #290, opened 2023-04-05
guide: Fixes calico hostprocess image version

claudiubelu:
**Reason for PR**:

In the guide, the calico.yml file is downloaded and the ``CALICO_VERSION`` is supposed to be replaced through sed. However, variable substitution doesn't work if it's put in apostrophes, which results in pods not working properly.


marosset (approved):
/lgtm
/approve

Issue #277, opened 2023-01-12
kube-flannel for linux fails to (re)start after rbac for kube-flannel for windows is applied


Mik4sa:
**Describe the bug**
Following [this guide](https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/guides/guide-for-adding-windows-node.md) for Kubernetes 1.26.0 the kube-flannel pod for linux is failing to start after a restart of the control plane (where the pod was running on). I no longer have the actual error message but the problem was that the service account used didn't have enough permissions.

I already analyzed that. The reason is the rbac file we need to apply at last:
> 5. Apply kube-flannel-rbac.yml from sig-windows-tools/kubeadm/flannel
> Next you will need to apply the configuration that allows flannel to spawn pods and keep them running:
> 
> ```bash
> kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/sig-windows-tools/master/hostprocess/flannel/flanneld/kube-flannel-rbac.yml
> ```

This one overwrites the `ClusterRole` and `ClusterRoleBinding` created by:
> 3. Apply the Flannel manifest and validate
> 
> Let's apply the Flannel configuration:
> 
> ```bash
> kubectl apply -f kube-flannel.yml
> ```
___
To overcome this I used a customized rbac file. This one works around the described issue:
``` yaml
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: flannel
  namespace: kube-system
---
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: flannel-windows
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: flannel
subjects:
  - kind: ServiceAccount
    name: flannel
    namespace: kube-system
---
kind: ConfigMap
apiVersion: v1
metadata:
  name: kube-flannel-cfg
  namespace: kube-system
  labels:
    tier: node
    app: flannel
data:
  cni-conf.json: |
    {
      "name": "cbr0",
      "plugins": [
        {
          "type": "flannel",
          "delegate": {
            "hairpinMode": true,
            "isDefaultGateway": true
          }
        },
        {
          "type": "portmap",
          "capabilities": {
            "portMappings": true
          }
        }
      ]
    }
  net-conf.json: |
    {
      "Network": "10.244.0.0/16",
      "Backend": {
        "Type": "vxlan"
      }
    }

```

[...]

PR #297, opened 2023-04-21
Support containerd 1.7

### LINKED ISSUE #289: PermissionDenied: (C:\hpc\etc\kube-flannel\net-conf.json:FileInfo) [Copy-Item], UnauthorizedAccessException
I think you need to know this https://github.com/flannel-io/flannel/issues/1758

Mik4sa:
**Reason for PR**:

The `kube-flannel` and `kube-proxy` pods were failing/crashing with containerd 1.7.0. This was because of the usage of `bind volume` instead of `symlink volume` (see [here](https://github.com/kubernetes/enhancements/tree/master/keps/sig-windows/1981-windows-privileged-container-support#compatibility)).
This problem was first mentioned in #289 and here: https://github.com/kubernetes-sigs/sig-windows-tools/issues/277#issuecomment-1512099019

For details check this comment: https://github.com/kubernetes-sigs/sig-windows-tools/issues/277#issuecomment-1518117147

**Issue Fixed**:

Fixes #289

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests

**Notes**:
I decided to move the mounts to `/mounts/...` instead of directly mounting them to the root because of two reasons:
- This didn't worked for kube-proxy since the Dockerfile already writes into/uses `/kube-proxy`
- `/mounts/...` maybe better communicates that these are actual mounts which was really hard for me to understand due to the behavior of hostprocess containers.

### DIFF
--- a/hostprocess/flannel/flanneld/start.ps1
+++ b/hostprocess/flannel/flanneld/start.ps1
@@ -8,15 +8,15 @@ Copy-Item -Path "$env:CONTAINER_SANDBOX_MOUNT_POINT/cni/*" -Destination "$env:CN
 Write-Host "copy flannel config"
 mkdir -force C:\etc\kube-flannel\
 ls C:\etc\kube-flannel\
-ls $env:CONTAINER_SANDBOX_MOUNT_POINT/etc/kube-flannel/
-cp -force $env:CONTAINER_SANDBOX_MOUNT_POINT/etc/kube-flannel/net-conf.json  C:\etc\kube-flannel\net-conf.json
+ls $env:CONTAINER_SANDBOX_MOUNT_POINT/mounts/kube-flannel/
+cp -force $env:CONTAINER_SANDBOX_MOUNT_POINT/mounts/kube-flannel/net-conf.json  C:\etc\kube-flannel\net-conf.json
 
 # configure cni
 # get info
 Write-Host "update cni config"
-$cniJson = get-content $env:CONTAINER_SANDBOX_MOUNT_POINT/etc/kube-flannel-windows/cni-conf-containerd.json | ConvertFrom-Json
-$serviceSubnet = get-content $env:CONTAINER_SANDBOX_MOUNT_POINT/etc/kubeadm-config/ClusterConfiguration | ForEach-Object -Process {if($_.Contains("serviceSubnet:")) {$_.Trim().Split()[1]}}
-$podSubnet = get-content $env:CONTAINER_SANDBOX_MOUNT_POINT/etc/kubeadm-config/ClusterConfiguration | ForEach-Object -Process {if($_.Contains("podSubnet:")) {$_.Trim().Split()[1]}}
+$cniJson = get-content $env:CONTAINER_SANDBOX_MOUNT_POINT/mounts/kube-flannel-windows/cni-conf-containerd.json | ConvertFrom-Json
+$serviceSubnet = get-content $env:CONTAINER_SANDBOX_MOUNT_POINT/mounts/kubeadm-config/ClusterConfiguration | ForEach-Object -Process {if($_.Contains("serviceSubnet:")) {$_.Trim().Split()[1]}}
+$podSubnet = get-content $env:CONTAINER_SANDBOX_MOUNT_POINT/mounts/kubeadm-config/ClusterConfiguration | ForEach-Object -Process {if($_.Contains("podSubnet:")) {$_.Trim().Split()[1]}}
 $na = Get-NetRoute | Where { $_.DestinationPrefix -eq '0.0.0.0/0' } | Select-Object -Property ifIndex
 $managementIP = (Get-NetIPAddress -ifIndex $na[0].ifIndex -AddressFamily IPv4).IPAddress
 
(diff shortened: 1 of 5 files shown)


jsturtevant:
> I decided to move the mounts to /mounts/... instead of directly mounting them to the root because of two reasons:

I don't think the `volumeMount:mountPath` bind directly to the `c:/folder` on host.  I would expect them to show up in `c:/hpc/`.  @marosset can help clarify.

It feels like something else is going on here? does the `C:\etc\kube-flannel\` folder exist on host?

jsturtevant:
That was my expectation, was the Working dir changed.  I am surprised that change the folder name is making a difference here.

Mik4sa:
@marosset What is missing here so this gets the lgtm label/gets merged?

hostprocess/flannel/flanneld/start.ps1
```diff
@@ -8,15 +8,15 @@ Copy-Item -Path "$env:CONTAINER_SANDBOX_MOUNT_POINT/cni/*" -Destination "$env:CN
 Write-Host "copy flannel config"
 mkdir -force C:\etc\kube-flannel\
 ls C:\etc\kube-flannel\
-ls $env:CONTAINER_SANDBOX_MOUNT_POINT/etc/kube-flannel/
-cp -force $env:CONTAINER_SANDBOX_MOUNT_POINT/etc/kube-flannel/net-conf.json  C:\etc\kube-flannel\net-conf.json
+ls $env:CONTAINER_SANDBOX_MOUNT_POINT/mounts/kube-flannel/
+cp -force $env:CONTAINER_SANDBOX_MOUNT_POINT/mounts/kube-flannel/net-conf.json  C:\etc\kube-flannel\net-conf.json
```
jsturtevant, line 12: What is the value of `CONTAINER_SANDBOX_MOUNT_POINT` in this script with 1.7? I am guessing we are getting the error `cp : Access to the path 'C:\etc\kube-flannel\net-conf.json' is denied` because we are trying to copy from the wrong place. If `CONTAINER_SANDBOX_MOUNT_POINT` is not set then we would be trying to copy from `c:/etc/kube-flannel/net-conf.json` which isn't created yet (we are trying to do that now...).

Mik4sa: Afaik 'C:/hpc' but I can't verify before wednesday/thursday

Mik4sa: > What is the value of `CONTAINER_SANDBOX_MOUNT_POINT` in this script with 1.7?

It's `C:\hpc\`. Just verified.

> I am guessing we are getting the error `cp : Access to the path 'C:\etc\kube-flannel\net-conf.json' is denied` because we are trying to copy from the wrong place.

I manually entered into the flannel pod and executed the commands one by one and made some other testing.
The source doesn't seem to be problematic. I'm able to copy it to a different folder. This for example works (after creating the target directory):
`cp -force $env:CONTAINER_SANDBOX_MOUNT_POINT/etc/kube-flannel/net-conf.json C:\test\`

But when I for example try to delete the target file with this command:
`Remove-Item C:\etc\kube-flannel\net-conf.json`
I get the following error:
```
Remove-Item : Cannot remove item C:\etc\kube-flannel\net-conf.json: Access to the path 'C:\etc\kube-flannel\net-conf.json' is denied.
At line:1 char:1
+ Remove-Item C:\etc\kube-flannel\net-conf.json
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : PermissionDenied: (C:\etc\kube-flannel\net-conf.json:FileInfo) [Remove-Item], UnauthorizedAccessException
    + FullyQualifiedErrorId : RemoveFileSystemItemUnAuthorizedAccess,Microsoft.PowerShell.Commands.RemoveItemCommand
```

This is because the configmap `kube-flannel-cfg` is mounted into this place (readonly). I verified that by editing the configmap first and then checking the content inside of the container:
```
PS C:\> cat C:\etc\kube-flannel\net-conf.json
{
  "Network": "10.244.0.0/16",
  "Backend": {
    "Type": "vxlan"
  }
  "Some falsy value..."
}
```

---

And this is the target directory structure:
```
PS C:\> ls C:\etc\kube-flannel\


    Directory: C:\etc\kube-flannel


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----         5/12/2023   2:54 PM                ..2023_05_12_14_54_31.2595385805
d----l         5/12/2023   2:54 PM                ..data
-a---l         5/12/2023   2:54 PM              0 cni-conf.json
-a---l         5/12/2023   2:54 PM              0 net-conf.json
```

jsturtevant: Ah, I yes this has to do with the behavioral change with mounts in containerd 1.7 (which I think you already said.  Thanks for bearing with me! 😄). 

from https://github.com/kubernetes/enhancements/tree/master/keps/sig-windows/1981-windows-privileged-container-support#container-mounts

> Additional volume mounts specified for hostProcess containers will be mounted at their requested location and can be access the same way as volume mounts in Linux or regular Windows Server containers.
ex: a volume with a mountPath of /var/run/secrets/token for containers will be mounted at c:\var\run\secrets\token for containers.

PR #301, opened 2023-05-03
Correct vxlan settings in guide

UweErikMartin:
Correction in section 5 - Apply kube-flannel-rbac.yml from sig-windows-tools/kubeadm/flannel

The file kube-flannel-rbac.yml contains also a net-conf.json section which must also
contain the VNI (4096) and the Port (4789). Added an sed-command to apply those
changes during installation.


UweErikMartin:
Reopen pull request after signing CLA

Issue #310, opened 2023-05-16
Pod calico-node-windows error ImagePullBackOff


kkbruce:
**Describe the bug**

Follow the document https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/guides/guide-for-adding-windows-node.md#configuring-calico-hostprocess ((Step 3) and complete Windows join the cluster. When the Windows node is ready, see the error from the console.

```
~$ kubectl describe pods calico-node-windows-4ccfb --namespace kube-system
Name:             calico-node-windows-4ccfb
Namespace:        kube-system
Priority:         0
Service Account:  calico-node
Node:             win-ubk1no1mgrs/192.168.58.130
Start Time:       Tue, 16 May 2023 14:12:07 +0000
Labels:           app=calico
                  controller-revision-hash=764bdf9f7c
                  pod-template-generation=1
                  tier=node
Annotations:      <none>
Status:           Pending
IP:               192.168.8.151
IPs:
  IP:           192.168.8.151
Controlled By:  DaemonSet/calico-node-windows
Init Containers:
  install-cni:
    Container ID:
    Image:         sigwindowstools/calico-install:v3.25.1-hostprocess
    Image ID:
    Port:          <none>
    Host Port:     <none>
    Args:
      $env:CONTAINER_SANDBOX_MOUNT_POINT/calico/install.ps1
    State:          Waiting
      Reason:       ImagePullBackOff
    Ready:          False
    Restart Count:  0
    Environment:
      CNI_CONF_NAME:         10-calico.conflist
      CNI_NETWORK_CONFIG:    <set to the key 'cni_network_config' of config map 'calico-config-windows'>  Optional: false
      KUBERNETES_NODE_NAME:   (v1:spec.nodeName)
      CNI_MTU:               <set to the key 'veth_mtu' of config map 'calico-config-windows'>  Optional: false
      SLEEP:                 false
      K8S_SERVICE_CIDR:      10.96.0.0/12
    Mounts:
      /etc/kubeadm-config/ from kubeadm-config (rw)
      /host/etc/cni/net.d from cni-net-dir (rw)
      /host/opt/cni/bin from cni-bin-dir (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-77867 (ro)
Containers:
  calico-node-startup:
    Container ID:
    Image:         sigwindowstools/calico-node:v3.25.1-hostprocess
    Image ID:
    Port:          <none>
[...]

PR #312, opened 2023-05-17
Add explanation for parameter $CALICO_VERSION

kkbruce:
**Reason for PR**:

The $CALICO_VERSION parameter does not refer to the same version value in the context, which can cause errors after the script runs. Add an explanation to avoid errors.

**Issue Fixed**:

Issue #310 

**Requirements**

- [ ] Sqaush commits 
- [x] Documentation
- [ ] Tests

**Notes**:


kkbruce:
Thank you for your review. Your meaning is more precise than what I expressed.

jsturtevant:
Thank you for taking the time to improve the docs!

/lgtm
/approve

PR #314, opened 2023-05-17
Remove custom kube-flannel-rbac

Mik4sa:
**Reason for PR**:
As described in #277 and #313 we should move kube-flannel for windows into the namespace `kube-flannel` so we can get rid of a custom rbac.
For now this is a draft because there is a change needed in flannel (see https://github.com/flannel-io/flannel/issues/1772) before we can go for it. Everything else needed should already have been addressed by this draft.

I discussed the needed changes with @jsturtevant back then (see https://github.com/kubernetes-sigs/sig-windows-tools/issues/277#issuecomment-1511680173).

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests


Mik4sa:
@jsturtevant Got it!
Check here for the last changes: https://github.com/kubernetes-sigs/sig-windows-tools/compare/45ad4057f58c704f60cf39cb896dc1255e39c2e6..bab7222c070469ac735b9f1d4d58ae1c3c839cc9

Basically we are directly communicating to the api server again but without using the kube-config as before.

So we don't need any changes in flannel or kube-proxy itself.

Mik4sa:
I had to make one more change:
https://github.com/kubernetes-sigs/sig-windows-tools/compare/bab7222c070469ac735b9f1d4d58ae1c3c839cc9..f0670fe7be356bfd1802beee58aff2d685a3d84b

I tried to get the `controlPlaneEndpoint` with
```
kubectl get configmap -n kube-system kubeadm-config -o yaml | grep controlPlaneEndpoint: | sed 's/.*\: //g'
```

but the configmap `kubeadm-config` doesn't have this value if it wasn't provided with the cli flag `--control-plane-endpoint` when initializing the cluster with `kubeadm init`.

So instead I get this value from the kube-proxy configmap with:
```
kubectl get configmap -n kube-system kube-proxy -o jsonpath="{.data['kubeconfig\.conf']}" | grep server: | sed 's/.*\:\/\///g'
```

Mik4sa:
Hey @jsturtevant did you already had time to check this out?

hostprocess/flannel/flanneld/start.ps1
```diff
@@ -32,13 +32,15 @@ Set-Content -Path $env:CNI_CONFIG_PATH/10-flannel.conf ($cniJson | ConvertTo-Jso
 Write-Host "add route"
 route /p add 169.254.169.254 mask 255.255.255.255 0.0.0.0
 
-write-host "copy sa info (should be able to do this with a change to go client"
-mkdir -force $env:CONTAINER_SANDBOX_MOUNT_POINT/flannel-config-file/var/run/secrets/kubernetes.io/serviceaccount/
-cp -force $env:CONTAINER_SANDBOX_MOUNT_POINT/var/run/secrets/kubernetes.io/serviceaccount/* $env:CONTAINER_SANDBOX_MOUNT_POINT/flannel-config-file/var/run/secrets/kubernetes.io/serviceaccount/
+# set host and port to Kubernetes API server
+$env:KUBERNETES_SERVICE_HOST = $env:CONTROL_PLANE_ENDPOINT.Split(':')[0]
```
jsturtevant, line 36: Is this going to override the ones created by kubernetes? https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables

hostprocess/flannel/flanneld/flannel-overlay.yml
```diff
@@ -126,15 +126,9 @@ spec:
               apiVersion: v1
               fieldPath: metadata.namespace
       volumes:
-      - name: kube-proxy
```
jsturtevant, line 129: this is great!

hostprocess/flannel/flanneld/flannel-overlay.yml
```diff
@@ -126,15 +126,9 @@ spec:
...
-        configMap:
-          name: kube-proxy
       - name: flannel-cfg
         configMap:
           name: kube-flannel-cfg
       - name: flannel-windows-cfg
         configMap:
           name: kube-flannel-windows-cfg
-      - name: kubeadm-config
```
jsturtevant, line 138: this is great!

hostprocess/flannel/flanneld/start.ps1
Mik4sa: This is replacing these envs
https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/#directly-accessing-the-rest-api

These envs are used by the in-cluster config. But instead going over the service created by Kubernetes we are directly communicating to the node (with it's actual IP)

ohanedan:
Hello,
Today, I updated flannel in my cluster from 0.14.0 to 0.22.0 and I can't use my windows node no longer. Even if I changed the namespace, I couldn't find solution because volume mount is looking for some resources in kube-system. I guess I'll have to wait for this mr's merge. Do we have any workaround solution?
Thanks.

Mik4sa:
> Hello @Mik4sa , Today, I updated flannel in my cluster from 0.14.0 to 0.22.0 and I can't use my windows node no longer. Even if I changed the namespace, I couldn't find solution because volume mount is looking for some resources in kube-system. I guess I'll have to wait for this mr's merge. Do we have any workaround solution? Thanks.

This PR isn't required to run an updated version of flannel. I'm running 0.21.5 successfully without this changes.
Can you open a new issue can ask for help there? Also please add some error messages you get.

Mik4sa:
@jsturtevant Is there anything missing to get this merged?

hostprocess/flannel/flanneld/start.ps1
jsturtevant: Sorry, I haven't had a chance to dig into this.  This seems strange to to overwrite what kubernetes is giving us.  I don't believe we needed to overwrite this information before.  I will try to get sometime soon to look more into this

Mik4sa: Oh well, I understand. I‘m sorry for bothering you
This is true. We weren‘t forced to overwrite this value before but the resulting behavior should be equal. Before we talked directly with one of the master nodes. With the overwrite we would instead talk over the service with one of the master node. This isn’t possible yet, though. Because the network isn’t ready for it yet. We discussed that a while back if you remember.
So instead of a larger change in flannel itself we are forcing Kubernetes (the pod) to talk to one of the master node directly by overwriting this value.

Don’t get me wrong. This isn’t the final solution. We’ll need that change in flannel. Especially if we want to integrate that more into the standard / kubeadm. But this isn’t possible by now

Though, the benefit for now is that we dropped dependencies to kubeadm-config and kube-proxy config 
In my opinion this is worth it. Also the needed change for us, once flannel made a change there, is way smaller if we merge this now. We simply update flannel and drop the overwriting :)

jsturtevant: not a bother! I really appreciate all the work you have done! I just haven't had chance to think deeply about it and with this type of change I want to make sure I fully understand.  The context you just added helps.

jsturtevant: ok, so it seems like we never ran into this because we _always_ had a kubeconfig with the management IP not service name (since kubeproxy has created all this). This was even the case when we installed flannel as a service on the host, not in a container.  I didn't look into the Linux side of things but they must be doing something (either in different order as you suggested or DNS is configured on the node?) different.  Since I don't think anyone's done this before it could be a bug in flannel as you suggested.

To unblock us, I think just setting the `KUBERNETES_SERVICE_HOST` env on the pod spec would be sufficient and not doing a conversion via script should work, could you try that?  

You have the open issue in flannel, might need someone to dive into Linux flannel pod and figure out what is going on there, so it could be solved long term.

Mik4sa: > [...] and not doing a conversion via script should work, could you try that?

Can you explain that a bit more? I'm unsure what you want me to do/try.

In case you mean that I should try to set the env inside of the yaml file: this could be problematic since I need to determine the value.

jsturtevant: > > [...] and not doing a conversion via script should work, could you try that?
> 
> Can you explain that a bit more? I'm unsure what you want me to do/try.
> 
> Do you mean that I should try to set the envs for host and port directly inside of the yaml? (why I didn't do this in the first place though 🤷‍♂️?)

yes, does that approach work?

Mik4sa: > > > [...] and not doing a conversion via script should work, could you try that?
> > 
> > 
> > Can you explain that a bit more? I'm unsure what you want me to do/try.
> > Do you mean that I should try to set the envs for host and port directly inside of the yaml? (why I didn't do this in the first place though 🤷‍♂️?)
> 
> yes, does that approach work?

That approach does indeed work, yes. Just tested and pushed the changes

jsturtevant:
could you a comment pointing at the flannel issue so it doesn't get lost in github PR? otherwise lgtm

Mik4sa:
> could you a comment pointing at the flannel issue so it doesn't get lost in github PR? otherwise lgtm

Once https://github.com/flannel-io/flannel/issues/1772 is solved we should update flannel and remove the custom kubernetes host and port from flannel-overlay.yaml

Like this?
These issues are already linked though. I already referenced it in the first "comment"

jsturtevant:
Sorry wasn't clear, I meant in the YAML file where we override the value.  I want to make sure if someone has that YAML they can get a link back to this issue or Flannel issue and understand why.  Once we close this PR it will be difficult for someone going through the Tutorial document to know why that value is overridden.

Mik4sa:
Yeah, this sound very reasonable. Will do so later the day or tomorrow :)

Mik4sa:
> Sorry wasn't clear, I meant in the YAML file where we override the value. I want to make sure if someone has that YAML they can get a link back to this issue or Flannel issue and understand why. Once we close this PR it will be difficult for someone going through the Tutorial document to know why that value is overridden.

Done.

Mik4sa:
@marosset I guess that you would approve this or not. Do you have that on your task list :)?

marosset:
> @marosset I guess that you would approve this or not. Do you have that on your task list :)?

sorry i missed this
/lgtm
/approve

Issue #317, opened 2023-05-18
Push new images to docker hub


Mik4sa:
**Describe the bug**
The images, atleast for flannel and kube-proxy, are outdated (see https://github.com/kubernetes-sigs/sig-windows-tools/pull/308#issuecomment-1551110003 and #316). We should push a new version to docker hub.

Also I feel like we should add an action to automatically push new Versions. Do you agree?

jonaskello:
I'm also looking for new flannel images but cannot find any.

Is the workaround to build them yourself for now?

Mik4sa:
See https://github.com/kubernetes-sigs/sig-windows-tools/issues/336#issuecomment-1633854019

> To build your own images prepare a docker installation powered by linux containers on a windows system. Execute this command to create the builder needed to build flannel and kube-proxy:
> ```
> docker buildx create --name img-builder --use --platform windows/amd64
> ```
> 
> Afterwars clone this repo and checkout the master branch. Open this folder in a terminal `/hostprocess/flannel/flanneld`.
> Set `$flannelVersion` and `$repository` accordingly. I used for example `$flannelVersion="v0.21.5"` and `$repository="mik4sa"`. Last but not least execute this command to build and push your image:
> ```
> docker buildx build --provenance=false --sbom=false --platform windows/amd64 --output=type=registry --pull --build-arg=flannelVersion=$flannelVersion -f Dockerfile -t $repository/flannel:$flannelVersion-hostprocess .
> ```
> 
> Next open this folder in a terminal `/hostprocess/flannel/kube-proxy`.
> Set `$proxyVersion` and `$repository` accordingly. I used for example `$proxyVersion="v1.27.1"` and `$repository="mik4sa"`. Last but not least execute this command to build and push your image:
> ```
> docker buildx build --provenance=false --sbom=false --platform windows/amd64 --output=type=registry --pull --build-arg=k8sVersion=$proxyVersion -f Dockerfile -t $repository/kube-proxy:$proxyVersion-flannel-hostprocess .
> ```
> 
> In the manifest you apply, so `flannel-overlay.yml` and `kube-proxy.yml` you need to reference the image you just created. So change `image: sigwindowstools/[...]` to `image: mik4sa/[...]`

PR #318, opened 2023-05-18
Adding GH action workflows for building flannel and flannel kube-proxy images

marosset:
**Reason for PR**:

- Adds a GH action for building new versions of the flannel image
- Updates the GH action used to build kube-proxy images to also build them for flannel CNI (instead of just calico)

Adding GH 
**Issue Fixed**:

Partially fixies #317 

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests

**Notes**:

/assign @Mik4sa

### DIFF
--- a/.github/workflows/build-flannel-images.yml
+++ b/.github/workflows/build-flannel-images.yml
@@ -1,0 +1,19 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+name: Build Flannel images
+
+on:
+  workflow_dispatch:
+    inputs:
+      flannel_version:
+        description: 'Version of flannel to build image for (ex: v0.21.5)'
+        required: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build and push images
+      run: |
+        echo "${{ secrets.DOCKER_SECRET }}" | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
+        cd ./hostprocess/flannel && ./build.sh -f ${{ github.event.inputs.flannel_version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build-flannel-images.yml
```diff
@@ -0,0 +1,19 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+name: Build Flannel images
+
+on:
+  workflow_dispatch:
```
Mik4sa, line 5: Should we add any other triggers?
What about on push to the default branch?

Am I right that workflow_dispatch means that it must be triggered manually?
If so, when will we trigger a manually build?

.github/workflows/build-flannel-images.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build and push images
+      run: |
+        echo "${{ secrets.DOCKER_SECRET }}" | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
+        cd ./hostprocess/calico && ./build.sh -f ${{ github.event.inputs.flannel_version }}
```
Mik4sa, line 19: Shouldn't the directory be "flannel" instead of "calico"?

marosset: yes -_-

.github/workflows/build-flannel-images.yml
marosset: Yes `workflow_dispatch` means triggered manually.

If we push on default branch we'd need another versioning strategy.
Right now the flannel version is used for the container image tag / version too and I think should avoid changing the contents of images we already pushed. If we push every time there are updates the default branch (even if we limit the job to changes under hostprocess/flannel) we would be updating the flannel images in between flannel releases.

Let's discuss triggers in a separate issue?

Mik4sa: Let's do so :)

Mik4sa:
@marosset Otherwise the changes looks good and we could merge IMO. Looks like I can't do much here sadly 😅

jsturtevant:
@Mik4sa thanks for the review!
/lgtm

PR #321, opened 2023-05-25
Add option to override the hostname for kubeadm

djboris9:
**Reason for PR**:
The `hostname` command returns the local node name whereas people might wish to use the FQDN or some other alternative domain name for the kubelet. This PR allows the user to specify an overridden hostname for kubeadm while having the default still set to the value returned by `hostname`.

**Issue Fixed**: No issue created

**Requirements**

- [X] Sqaush commits 
- [ ] Documentation
- [ ] Tests

**Notes**:


jsturtevant:
I am going to approve this but please note that hostnames that are longer than 15 characters can have compatibility issues: https://github.com/kubernetes-sigs/cluster-api/issues/2217

/approve

PR #325, opened 2023-05-31
Improvements to InstallContainerd.ps1 and PrepareNode.ps1

coutinhop:
(from the Calico forks on https://docs.tigera.io/calico/3.26/scripts/PrepareNode.ps1 and https://docs.tigera.io/calico/3.26/scripts/Install-Containerd.ps1)

InstallContainerd.ps1:
- Add $CNIBinPath and $CNIConfigPath optional flags (defaulting to the currently hardcoded "c:/opt/cni/bin" and "c:/etc/cni/net.d")
- Make sure both paths exist with 'mkdir -force'

PrepareNode.ps1:
- Add setup for log rotation for kubelet on nssm

**Reason for PR**:

**Issue Fixed**:

Issue #

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests

**Notes**:

### DIFF
--- a/hostprocess/Install-Containerd.ps1
+++ b/hostprocess/Install-Containerd.ps1
@@ -19,8 +19,14 @@ Name of network adapter to use when configuring basic nat network.
 Skip the CPU check for Hypervisor support. You way wont be able to host Hyper-V isolated containers.
 Check https://github.com/kubernetes-sigs/sig-windows-tools/issues/296#issuecomment-1511695392 for more information.
 
+.PARAMETER CNIBinPath
+Path to configure ContainerD to look for CNI binaries.
+
+.PARAMETER CNIConfigPath
+Path to configure ContainerD to look for CNI config files.
+
 .EXAMPLE
-PS> .\Install-Conatinerd.ps1 -ContainerDVersion 1.7.1 -netAdapterName Ethernet -skipHypervisorSupportCheck
+PS> .\Install-Containerd.ps1 -ContainerDVersion 1.7.1 -netAdapterName Ethernet -skipHypervisorSupportCheck -CNIBinPath "c:/opt/cni/bin" -CNIConfigPath "c:/etc/cni/net.d"
 
 #>
 
@@ -31,6 +37,10 @@ Param(
     [string] $netAdapterName = "Ethernet",
     [parameter(HelpMessage = "Skip the CPU check for Hypervisor support. Note that you will not be able to host Hyper-V isolated containers")]
     [switch] $skipHypervisorSupportCheck
+    [parameter(HelpMessage = "Path to configure ContainerD to look for CNI binaries")]
+    [string] $CNIBinPath = "c:/opt/cni/bin"
+    [parameter(HelpMessage = "Path to configure ContainerD to look for CNI config files")]
+    [string] $CNIConfigPath = "c:/etc/cni/net.d",
 )
 
 $ErrorActionPreference = 'Stop'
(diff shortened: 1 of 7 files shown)


hostprocess/Install-Containerd.ps1
Mik4sa, line 1: Actually flannel is also dependent on these two paths:
https://github.com/kubernetes-sigs/sig-windows-tools/blob/d744f5488dd350fa967b3920d698b268e214e14d/hostprocess/flannel/flanneld/flannel-overlay.yml#L114-L117

We would need a change there too. Maybe we could just get these paths from config.toml directly? So we don't need to configure them twice?

I have no experience with calico. Could this also be a problem for calico?

Mik4sa (changes requested):

hostprocess/Install-Containerd.ps1
coutinhop: Calico HPC (currently tech-preview, GA coming soon) also needs these paths (currently needing them in a config-map with the same defaults as here, we're looking into autodetecting them). I believe we should be able to config the paths here at containerd installation, and maybe have flannel/calico/etc try to detect them? (tldr: I agree we shouldn't have to config them twice, but I believe here is the right place to do so). Please let me know your thoughts, thanks!

Mik4sa: I'm unsure what you mean. Can you give some more detail? Or an example?

coutinhop: Sure! I meant that calico uses `CNI_BIN_DIR` and `CNI_CONF_DIR` in the `calico-windows-config` configmap (documented in https://docs.tigera.io/calico/latest/getting-started/kubernetes/windows-calico/quickstart#install-calico-for-windows-using-hostprocess-containers). This is currently in tech-preview, we're working on improving calico HPC support and looking into a way to autodetect the paths.

I also meant that the `Install-Containerd.ps1` is the earliest point in time where we can configure these paths (since it's a prereq to installing kubernetes with HPC support, right?), so I believe making it configurable here is the right thing to do. I think calico, and flannel and other CNIs should try to detect these paths if possible. But I think changes to do so would not be in the `Install-Containerd.ps1` script. Additionally, I think it's a good thing to have this configurable, and optional with a sane default. As it is, if the user doesn't specify those args to `Install-Containerd.ps1`, everything stays the same.

(sorry, I'm trying to make a point for the changes in this PR, apologies if I'm mistaking your comments as disagreeing they should happen)

Mik4sa: Here you find the guide(s):
https://github.com/kubernetes-sigs/sig-windows-tools/tree/d744f5488dd350fa967b3920d698b268e214e14d/guides

Where "guide-for-adding-windows-node.md" is the "main" guide. You may add a note in here, but maybe also in "calico.md" and "flannel.md"

jsturtevant:
the changes lgtm,  @coutinhop did you want to update the docs to be clearer about the setup?

coutinhop:
> the changes lgtm,  @coutinhop did you want to update the docs to be clearer about the setup? 

Yes, will do! Apologies, had some personal issues keeping me busy, will do so right away

hostprocess/Install-Containerd.ps1
coutinhop: Done. These paths on the calico files are a little say, "less than ideal", but I believe the calico guide will need an overhaul once we (I'm on the project calico team) push hostprocess containers into GA. I'll be happy to make a PR with suggestions for updates once that's ready.
@Mik4sa @jsturtevant please take a look, and thanks for your patience so far!

Mik4sa: The latest changes according to the guide look good so far. Haven't tried it so far whether it works. I may get time on sunday for it to have a last look on it :)

coutinhop: @Mik4sa sorry to bother you, but did you have the opportunity to take a look at the latest changes here? Thanks!

Mik4sa: Not yet. I'm really sorry. Didn't find time for it. I'm starting just now

Mik4sa: @coutinhop Looks good to me. Everything worked for me.
For reference, I used `C:/Program Files/containerd/cni/bin` and `C:/Program Files/containerd/cni/conf`

PR #329, opened 2023-06-13
use the windows-host-process-containers-base-image for minimal image size

davhdavh:
**Reason for PR**:
Change base image to windows-host-process-containers-base-image for minimal image size and compat with all windows versions in 1 image


hostprocess/csi-proxy/build.sh
```diff
@@ -60,12 +53,6 @@ fi
 # Create manifest
 docker manifest create ${repository}/csi-proxy:${version} $manifest_entries
 
-# Annotate manifests
-for win_ver in "${win_vers[@]}"; do
```
jsturtevant, line 64: @marosset  do we still have an issue in containerd where this won't pull properly with out the right manifest or did that get resolved?

marosset:
I think these changes look good.
I will enlist in the PR branch and check things out this week.

marosset:
@davhdavh - Can you rebase this so it'll pick up some of the changes from https://github.com/kubernetes-sigs/sig-windows-tools/pull/331?

And also if this is switched to a host-process container we don't need to make or push a manifest so everything after the `# Create manifest` line in the build.sh script should be removed.

thanks!
--Mark

jgrieger:
Hi @jayunit100, would you be so kind and review this dangling PR? Would be much appreciated. 😊

davhdavh:
merge it?

jgrieger:
> merge it?

Seems that there is a label missing:
image

Idk, who applies this label.

davhdavh:
@jsturtevant I think it needs your approval

PR #352, opened 2023-12-20
docs: Consistent format of versions input for scripts

mloskot:
**Reason for PR**:

This make versions input a no-brainer as currently it may alert users with unnecessary questions "is that a typo? which one is correct? shall I use `v` prefix for the ContainerD installer?", etc.

The `PrepareNode.ps1` can handle both formats `X.Y.Z` and `vX.Y.Z` though, but one has to read the script to learn it.

**Issue Fixed**:

Issue #

**Requirements**

- [ ] Sqaush commits 
- [ ] Documentation
- [ ] Tests

**Notes**:


marosset:
> This make versions input a no-brainer as currently it may alert users with unnecessary questions "is that a typo? which one is correct? shall I use v prefix for the ContainerD installer?", etc.

I'm in favor of using the `v` prefix everywhere personally

mloskot:
Yes, I have realised `v` prefix is de-facto a standard convention used across Kubernetes tooling.

PR #371, opened 2024-07-08
Add Calico Windows Guide for Install using the operator

fabi200123:
**Reason for PR**:
Since Calico version v3.27.0, it was implemented another way of installing calico on Windows Nodes, using the [operator](https://docs.tigera.io/calico/latest/getting-started/kubernetes/windows-calico/operator).

For the newer versions of Calico (v3.27, v3.28), the `calico-hostprocess` images are missing, thus resulting in errors if someone follows the previous guides.

### DIFF
--- a/guides/calico_operator.md
+++ b/guides/calico_operator.md
@@ -1,0 +1,66 @@
+# Installing Calico using the Operator
+
+This should be followed after adding your [Windows node to the cluster with kubeadm](guide-for-adding-windows-node.md#adding-windows-nodes).
+
+>  **Note** This guide is available ONLY for Calico version greater or equal to **v3.27.0** and for Kubernetes versions that support **HostProcess containers (HPC)**.
+
+## Configuring Calico with HostProcess containers using the Operator
+
+>  **Note** All code snippets in Linux sections are to be run in a Linux environment on the Linux worker node.
+
+### Prepare Control Plane for Calico
+
+1. Prepare Kubernetes control plane for Calico
+
+Some minor preparation is recommended on the Kubernetes control plane in our cluster. It is recommended to enable bridged IPv4 traffic to iptables chains when using Calico. The following command must be run on all Linux nodes:
+
+```bash
+sudo sysctl net.bridge.bridge-nf-call-iptables=1
+```
+
+2. Download & configure Calico for Linux
+
+The version of Calico will be set in the variable CALICO_VERSION (if another version is needed, modify this variable).
+
+```bash
+export CALICO_VERSION="v3.28.0"
+```
+
+Create the tigera-operator.yaml and the custom-resources.yaml for Calico:
+```bash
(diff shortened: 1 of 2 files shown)


guides/guide-for-adding-windows-node.md
```diff
@@ -65,3 +65,4 @@ This guide offers two choices:
```
coutinhop, line 65: nit about the line above: should become "This guide offers **three** choices:"

fabi200123: Thanks, updated it!

guides/calico_operator.md
```diff
@@ -0,0 +1,66 @@
...
+- Modify the IPPool to have `vxlanMode: Always`
+
+```bash
+kubectl patch ippool default-ipv4-ippool --type='json' -p='[{"op": "replace", "path": "/spec/vxlanMode", "value": "Always"}]'
+```
+
+- Modify VXLAN for `installation` default's resource from `VXLANCrossSubnet` to `VXLAN`
+
+```bash
+kubectl patch installation default --type='json' -p='[{"op": "replace", "path": "/spec/calicoNetwork/ipPools/0/encapsulation", "value": "VXLAN"}]'
```
coutinhop, line 62: Patching the IP pool shouldn't be necessary if you install calico from the start (i.e. linux) with `encapsulation: VXLAN` (not cross subnet) in the `installation` resource. I'll work on updating the Calico docs to make that more explicit.

Also, VXLAN is not strictly necessary, as Calico for Windows also supports no encapsulation (`encapsulation: None`) with BGP enabled. The Calico docs do have more details on how to set up on both scenarios.

guides/calico_operator.md
```diff
@@ -0,0 +1,66 @@
...
+### Add Windows Calico Solution with the Operator
+
+For Linux control nodes using Calico networking, strict affinity must be set to true. This is required to prevent Linux nodes from borrowing IP addresses from Windows nodes:
+
+```bash
+kubectl patch ipamconfigurations default --type merge --patch='{"spec": {"strictAffinity": true}}'
+```
+
+>  **NOTE** If the above command failed to find `ipamconfigurations` resource, you need to install the Calico API server. Please refer to [installing the Calico API server](https://docs.tigera.io/calico/latest/operations/install-apiserver).
+
```
coutinhop, line 48: Good catch, I think the Calico docs should also mention this. Will add, thanks!

guides/calico_operator.md
fabi200123: I have mentioned this steps regarding the "Kubernetes VXLAN installation"
Do you think I should add a note regarding BGP with a link towards Calico documentation?

guides/calico_operator.md
fabi200123: I think this is also mentioned on the Calico part, but I am not entirely sure

PR #395, opened 2026-03-03
Synchronize automatically kube-proxy images

juanluisvaladas:
**Reason for PR**:

Currently the build-kube-proxy-images.yml workflow must be triggered manually, this PR automates that task so that it's made automatically for every new version. We only check versions starting on 1.33 because it's the oldest maintained branch.

This was tested here: https://github.com/juanluisvaladas/sig-windows-tools/actions/runs/22623814294/job/65555151567#step:4:62

In this test we can see the images for 1.35 not being triggered because they existed from previous execution runs and we see the workflows dispatched for the missing images.

**Requirements**

- [x] Sqaush commits 
- [x] Documentation
- [x] Tests

**Notes**:

Tested manually and I don't think docs are required for this change.

### DIFF
--- a/.github/workflows/sync-kube-proxy-images.yaml
+++ b/.github/workflows/sync-kube-proxy-images.yaml
@@ -1,0 +1,92 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+name: Sync Kube-Proxy images
+
+on:
+  schedule:
+    - cron: '0 0 * * *' # Every day at 00:00 UTC
+  workflow_dispatch:
+    inputs:
+      repository:
+        description: 'Repository to push the image to (ex: docker.io/sigwindowstools)'
+        required: false
+        default: 'docker.io/sigwindowstools'
+
+env:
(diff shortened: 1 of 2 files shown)


.github/workflows/sync-kube-proxy-images.yaml
```diff
@@ -0,0 +1,92 @@
...
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        set -euo pipefail
+
+        REPOSITORY="${{ env.REPOSITORY }}"
+        MIN_VERSION="1.33"
+        SUFFIX="-calico-hostprocess"
+        MISSING=()
+
+        # Fetch all kube-proxy tags from registry.k8s.io
```
juanluisvaladas: This logic is a direct translation from test-kube-proxy-images.ps1

juanluisvaladas:
> @juanluisvaladas - Can you disable the cron schedule for https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/.github/workflows/test-kube-proxy-images.yaml too?
> 
> 
> 
> We can keep the workflow around for a few weeks but it shouldn't be needed with these changes!

Hey, sure I'm out of office but I'll do it next week.

PR #398, opened 2026-04-13
Sync kube-proxy pre-release versions

juanluisvaladas:
**Reason for PR**:
In k0s we'd like to have pre-release tags so that we can bump the kubernetes version before the final release.
This modifies the workflow to sync images so that the pre release images and also checks them in the job to validate they are synced.

**Issue Fixed**:
**Requirements**

- [x] Squash commits 
- [ ] Documentation (I don't think this is truly required)
- [x] Tests

**Notes**:

### DIFF
--- a/.github/workflows/sync-kube-proxy-images.yaml
+++ b/.github/workflows/sync-kube-proxy-images.yaml
@@ -51,11 +51,8 @@ jobs:
           # Must start with 'v'
           [[ "$TAG" != v* ]] && continue
 
-          # Skip pre-release tags (contain a hyphen after stripping 'v')
-          VER="${TAG#v}"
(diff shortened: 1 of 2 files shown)


aravindhp (commented):
In general I am not opposed to this. @marosset @zylxjtu will the increase in number of published images causes on the registry side?

juanluisvaladas:
Hi, Kubernetes is starting to do alpha releases for 1.37, we don't really need this yet but ideally we need to start thinking whether we build the images ourselves or if we can rely on this. Could we please get an update on this?

marosset:
it looks like only alpha/beta/rc tags would get added which shouldn't add too many images (after the initial sync)
/approve
/lgtm

